PR #71, opened 2016-04-10
Updating to Version 2

eirannejad:
Updating to version two:
Startup script now lives outside the scripts folder
pyRevit is now a multi-tab system. Startup script creates a tab for every folder in its home directory that includes scripts.
Startup script now works with both RevitPythonShell or RevitPythonLoader.


CHBIV007:
Can you post a photo or Screen shot of how it looks on your end?

eirannejad:
Working on a better documentation

PR #118, opened 2016-08-10
New Script: [Memory] Copy and Paste Viewport Placement

gtalarico:
@eirannejad this is the copy/paste view port placement script I mentioned to you.
feel free to modify as you wish, including header/notice, etc.


eirannejad:
Gui,Thanks so much for these. I added them to the repo and also added the
Author to the tooltips. I made major modifications to these so these tools
can convert coordinates between model and sheet space and position the
viewports based on the model space on the sheets. This is very important
for architects to have all the plan viewports aligned on multiple sheets
for coordination and overlay.

So I added my name to the author after yours. Let me know if this is okay
with u :)

Thanks a bunch again!

gtalarico:
Makes perfect sense.
Thanks for improving it!

Worked for me because I had all views cropped to the same scope box, so it
was kind of "hacky" :)

Gui Talarico
[ mobile ]

PR #124, opened 2016-09-11
Package Manager Improvements

gtalarico:
- Auto Center Window
- Shorten commit hash to 5 characters to save space
- Added Install+Update Labels to buttons (replaces Clone)
- Using pull -f instead of fecth/checkout when updating
- Sized columns more appropriately (not auto yet)


eirannejad:
Masterful as usual. Thanks pal

PR #213, opened 2017-07-12
Detail curves filter before making pattern

melnikovalex:
Hi Ehsan!

I tried pattern maker and it's absolutely great. 

But once I run it with model lines selected and it exception was risen on line 37 (it was trying to get OwnerViewId of model line).
I added a filter for selection before opening MakePatternWindow.. Now it leaves only detail lines.
Maybe you will find better sollution.

Anyway, thanks a lot!


eirannejad:
Very very nice! Thank you!! Merged to master.

PR #242, opened 2017-10-12
listFamilySizesCreatorGrouped - add

hdm-dt-fb:
as discussed: list project families and their sizes on disk (if found) grouped by creator


eirannejad:
Added to PR#241 since this is the continuation

eirannejad:
Merged to master

PR #262, opened 2017-12-06
Fixing installer error "must be logged as administrator"

melnikovalex:
Dear Ehsan, 

I was unable to run installer on computer without admin rights in my office. 

Then I figured out that one line in InnoSetup settings file resolves this issue. I tried to build an installer with it and it helped.

`PrivilegesRequired=lowest`

Hope it's matter. 

Thank you, 
Alex

### DIFF
--- a/release/setupmaker/pyRevitSetup.iss
+++ b/release/setupmaker/pyRevitSetup.iss
@@ -23,6 +23,7 @@ AppCopyright=Copyright (c) 2014-2017 Ehsan Iran-Nejad
 DisableWelcomePage=no
 WizardImageFile=installerimage.bmp
 LicenseFile=LICENSE
+PrivilegesRequired=lowest
 
 [Languages]
 Name: "english"; MessagesFile: "compiler:Default.isl"


eirannejad:
Well this is generally not accepted here since a person with no Admin rights is not supposed to install applications on their system. Thanks for the PR though.

eirannejad:
@melnikovalex Hey thanks for this info. I added it to the installer as comment. I need to talk to a few people and see if it makes sense to remove the admin requirement. Best practice says that a user with no admin rights should not be able to install applications. But I need to gather more info on this.

melnikovalex:
@eirannejad thank you. My opinion is that "All users / Only for me" option during install do the same. I mean, you can't install for All users without admin rights anyway

eirannejad:
@melnikovalex I completely agree. The problem is that if I change this now, companies that install pyRevit on the users and don't want the users to have admin privileges will freak out.

Let me think about this a bit more and discuss it with more people.

PR #284, opened 2018-01-24
Fixed unicode error on call to json.dumps in slackclient lib

tymokvo:
forgot this is a really easy fix. opened an issue on slackclient but who knows if they will accept it


eirannejad:
Merged. Thank you! @tymokvo

PR #332, opened 2018-04-30
add createFreeFormElement

hdm-dt-fb:
simple yet handy converter to get more flexibility out of family SAT imports.


eirannejad:
Will review this weekend. Thanks for the PR!

PR #340, opened 2018-06-05
Feature/locales - Tracking modified time

Darkkrynn:
Good morning,

We use PyRevit from time to time for some rather convenient function so who has to modify what.

We would like to know if it is possible to have the day and/or the time of the modification of the selected object.

Do you think it is possible to get this information from revit? And do you think it would be nice to be able to add it?

thanks
2018-06-06_01h26_35


eirannejad:
@Darkkrynn This is not the correct place for a question. You need to create an issue to email me. This is the Pull requests section for when you have made changes to pyRevit and would like to add them to the original repository for everyone else.

I'd appreciate if you can create an issue here: https://github.com/eirannejad/pyRevit/issues

To answer your question, it depends if revit keeps time stamp of changes. It didn't when I created that tool but I can take another look and see if maybe they have added that feature.

PR #358, opened 2018-07-26
Add a reset_config or clear_config method

CyrilWaechter:
Hi,
There is already a save_config and get_config method but if you want to set config back to default you need to reset each value in your script or manually delete section from config file.
What do you think about adding a method to reset config by deleting section from config file ?


eirannejad:
Very very handy! Thanks a lot @CyrilWaechter

PR #447, opened 2018-12-05
Master

thazell:
Had an error with single transactions, so I added a transaction group.  It runs slower, but was the only solution for that project.


thazell:
Looks like the Increment was fixed, but not the decrement.  I tried to fix the decrement as well, but not sure it got  caught for the pull request.

thazell:
>           Looks like the Increment was fixed, but not the decrement.  I tried to fix the decrement as well, but not sure it got  caught for the pull request.

Looks like it showed up!

eirannejad:
@thazell Thanks a lot for the pull request. I merged into dev branch and will push with 4.6.9 this weekend.

PR #455, opened 2018-12-21
 Changed matches() of TemplateListItem

Tomey77:
Split filter string and match individual words rather than the whole string


eirannejad:
Reviewing now...Thanks @Tomey77

eirannejad:
Added fuzzy search. will publish with 4.6.9
https://github.com/eirannejad/pyRevit/commit/71a7c7ec17df5fc72be63224303eaacf603e70a1

PR #478, opened 2019-01-14
Correcting WPFWindow.set_image_source doc string

CyrilWaechter:


eirannejad:
Thanks a lot @CyrilWaechter. Great catch and sorry for the confusion.

PR #548, opened 2019-03-24
Extended pyRevit run CLI behavior to automatically copy .addin files into the batch running context.

tatlin:
- Added code to copy all .addin files and subdirs into the batch running context. This is to workaround problems with 'built-in' Revit addins like the NavisExporter not being available while using pyRevit batch.
- This copy operation should be fairly robust and will only happen once while getting ready to run the batch.  The files should get cleaned up after the run
- This copy will only pick up addins installed in  AllUsers location (C:\ProgramData\Autodesk\Revit\Addins\2019) but could be generalized to handle CurrentUser too.
- I have tested this locally with Revit 2019 but it should work with older versions as well.


eirannejad:
Merged into dev and schedules for 4.7 release.

PR #582, opened 2019-04-28
Settings window UI; fix duplicated extentions paths

melnikovalex:
Hey Ehsan! There are couple of fixes, please have a look.

**1. UI Improvements of Settings window.**

If there is no write-access to pyRevit.addin file (if pyRevit deployed as admin), it locks iron python engine checkbox in Settings window.

**2. Remove duplicates from extentions folders list.** 
For me it was set in config.ini:

`userextensions = ["C:\\Users\\Alex\\AppData\\Roaming\\pyRevit\\Extensions"]`

and in `userconfig.py` there are lines:
```
if op.exists(EXTENSIONS_DEFAULT_DIR):
            dir_list.append(EXTENSIONS_DEFAULT_DIR)
```

which causes duplicated pathes as result of `get_ext_root_dirs()`. And finally all the extensions were reloaded two times on each reload.

I do not sure, but maybe it makes sence to remove `EXTENSIONS_DEFAULT_DIR` at all, because it being written into config from `pyrevitcli` anyway. When I install an extension from UI, it uses pyrevit cli, isn' it?

Thank you!

### DIFF
--- a/extensions/pyRevitCore.extension/pyRevit.tab/pyRevit.panel/Settings.smartbutton/script.py
+++ b/extensions/pyRevitCore.extension/pyRevit.tab/pyRevit.panel/Settings.smartbutton/script.py
@@ -145,6 +145,10 @@ class SettingsWindow(forms.WPFWindow):
                     self.availableEngines.SelectedItem = engine_cfg
                     break
 
+            # if addin-file is not writable, lock changing of the engine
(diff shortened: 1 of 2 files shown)


eirannejad:
Manually merged into develop. This is a great improvement. Thanks!

Issue #638, opened 2019-06-27
WPF example file error


sajusathish:
Hi I am new to WPF and wanted to explore the example file in the pyrevit documentation.
I have just copied the script that uses content and literal_string to create forms and I received the following error, am I missing something?

![image](https://user-images.githubusercontent.com/39821809/60266518-d17c9600-9905-11e9-950f-7fa2e8363959.png)

thanks.

eirannejad:
Watch a tutorial on WPF first. WPF is similar to XML and the namespaces must be defined otherwise the processor can not find the types. Your script is failing because the namespace that contains the type `Window` is not provided. See below:

```xaml
<Window
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    ShowInTaskbar="False" ResizeMode="NoResize"
    WindowStartupLocation="CenterScreen"
    HorizontalContentAlignment="Center">
</Window>
```

PR #642, opened 2019-06-29
Docs: Add namespaces in XAML example for WPFWindow

djsiroky:
Re: #638 

I ran into this issue today and saw someone else experienced the same thing recently, so hopefully this can prevent further issues created for you. Thanks for pyRevit! 🙏


eirannejad:
@djsiroky Thank you! I didn't even realize this is wrong on the docs.

PR #659, opened 2019-08-02
Scripts imporvements

melnikovalex:
7f91cb81@Revert "Select Reveals - select horizontal with shiftclick"
740efcae@Copy/Move viewports - do not show Sheet Placeholders in a list
cba850fe@Overkill - filter to leave DB.Line only to not to cause exception
729474a1@"List elements on selected levels" - ability to pick levels from list
ef1b5b5e@select_sheets(), select_views() - option to use current selection, fix
df894b51@DrawingSet scripts - use pre-selected views if possible
6e88cdf9@forms.select_sheets(), select_views() - option to use current selection


eirannejad:
♥ I love your PRs...So professional and thoughtful. Thank you!

PR #660, opened 2019-08-02
Scripts improvements

melnikovalex:
Hey Ehsan! I've just finished small bunch of changes for`pyRevitTools` scripts and couple of changes in `pyrevitlib`. Hope you'll find that useful:

**Fixes:**
1642b337 Overkill - there was an exception, in case if there are non-DB.Line curvers in a project (e.g. Arcs)

**pyrevitlib improvements**
2316ecb9 542f9c91 select_sheets(), select_views() - I've added an option to use preselected elements (ones which pass through filter) instead of selecting them anew. User will be asked if he wants to select from List or use preselected Elements. 
In similar way, there is an option to use Active view instead of selecting it in list.

7039233a I've added select_levels() method which works the same way with preselected elements.

**Scripts improvements**
296e008e DrawingSet scripts - use pre-selected views if possible (an implementation of changes described above)
7039233a "List elements on selected levels" - ability to pick levels from list
123c7595 "Copy/Move viewports" - do not show Sheet Placeholders in a list

53e71351 "Select (vertical) Reveals" - now you can select also horizontal ones, using shiftclick


eirannejad:
@melnikovalex Made a few adjustments and merged. Mostly simplified the forms module changes

PR #670, opened 2019-08-14
New tools - Save families, Select top groups, Keep mirrored

melnikovalex:
Hey Ehsan! 

There are 3 new tools which I would love to transfer to pyRevit Tools. They are pretty easy and wouldn't require any support:

- Save families
- Select top groups
- Keep mirrored

Sorry I couldn't manage to split the pull request from previous commits, 
there is only one you need d87bfd0


eirannejad:
@melnikovalex refactored code and merged
Thank you for the awesome new tools!!

PR #717, opened 2019-10-09
ReNumber tool - non-English Revit support

melnikovalex:
Hey Ehsan.

I've updated ReNumber tool so it can work with non-English Revit.
Basically, I've replaced all category names with BuiltInCategories.

Hope it will be helpful. Please, take a look


extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/edit2.stack/ReNumber.pushbutton/script.py
```diff
@@ -12,18 +12,22 @@
...
             target_view.EnableTemporaryViewPropertiesMode(target_view.Id)
-        rr_cat = revit.query.get_subcategory(category_name, 'Reference')
-        rr_cat.Visible[target_view] = state
-        rr_int = revit.query.get_subcategory(category_name, 'Interior Fill')
-        if not rr_int:
-            rr_int = revit.query.get_subcategory(category_name, 'Interior')
-        rr_int.Visible[target_view] = state
+        try:
+            # subcategories names (e.g. OST_MEPSpaceReferenceVisibility, reduce s)
+            bicat_reference = DB.BuiltInCategory[str(bicat)[:-1] + "ReferenceVisibility"]
```
melnikovalex: Do not sure if it's correct to take BIC like that, but it works. For emergeny cases I've put it into try/catch. Is there something like `.get(x, None)` for Enums?

eirannejad:
Reviewed and merged into develop

PR #722, opened 2019-10-14
"Copy/Paste state - Viewport position" tool improvements

melnikovalex:
Hey Ehsan, 
Hey Gui

I've got several feedbacks on "Copy/Paste state - Copy viewport position" tool from my colleagues.
So am trying to make it work better in certain cases (but the work is still in progress)

Therefore I think it would be great if @gtalarico would have a look as well.

Please, let me know if I can break something or if I shouldn't touch it at all.

- [x] support Sheets with different TitleBlock locations

- [x] hide viewport on a sheet when getting a cropbox

- [x] allow to choose ProjectBasePoint as origin (instead of Cropbox)

- [ ] show the option "By ProjectBasePoint" only if possible

- [ ] allow to choose cropbox alignment (top left, bottom left etc.) (optional, with __shiftclick__?)

- [ ] support Sections/Elevation views (by project base point)

- [ ] support active view (not selected viewport)

- [ ] support Pasting to multiple selected elements at once (relevant for all Copy/Paste modes)

Thank you too

### DIFF
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Selection.panel/memo.stack/lib/viewport_placement_utils.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Selection.panel/memo.stack/lib/viewport_placement_utils.py
@@ -58,15 +58,16 @@ def transform_by_matrix(coord, transmatrix):
     return DB.XYZ(newx, newy, 0.0)
 
 
-def set_tansform_matrix(selvp, selview, vpboundaryoffset, reverse=False):
+def set_tansform_matrix(selvp, selview, vpboundaryoffset=0, reverse=False):
+    # TODO support alignment by project basepoint, not by cropbox
+    # TODO support sections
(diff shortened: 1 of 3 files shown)


extensions/pyRevitTools.extension/pyRevit.tab/Selection.panel/memo.stack/lib/viewport_placement_utils.py
```diff
@@ -0,0 +1,179 @@
+from pyrevit import revit, script, forms, DB
```
melnikovalex, line 1: I've moved all methods which are used both in Copy and Paste scripts to separated script in `lib`

melnikovalex:
Closed, because I rewrote this in another way in #763

eirannejad:
@melnikovalex Thanks a lot. I'm going thru job change but will merge these soon

PR #731, opened 2019-10-25
Failure swallower context

melnikovalex:
Hey Ehsan!

Here I wanted to solve the problem of pop ups with warning while you're checking or exporting families.

I know there is an option `swallow_errors` for `revit.db.transaction`. I wanted to have something similar to apply without transaction. Here is a usage sample:

```
with FailureSwallowerContext(force=True) as swallower: 
   # `force` solves to resolve not only warnings but errors as well
   for fam in families:
      revit.doc.EditFamily(fam)
      if swallower.get_swallowed():
         logger.warn("Warnings swallowed")
```

I've also implemented it into "Families Save" and "QuickCheck" tools.

Please have a look. There are my comments in review

### DIFF
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/ptools.stack/Family.pulldown/Family QuickCheck.pushbutton/FamilyQuickcheck_script.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/ptools.stack/Family.pulldown/Family QuickCheck.pushbutton/FamilyQuickcheck_script.py
@@ -6,15 +6,24 @@ github.com/frederic-beaupere
 """
 #pylint: disable=invalid-name,import-error,superfluous-parens,broad-except
 
-from pyrevit import revit, DB
-from pyrevit import script
+from pyrevit import revit, script, DB, forms
+from pyrevit.revit import FailureSwallowerContext
 
 
 __author__ = 'Frederic Beaupere\n{{author}}'
 
-
 logger = script.get_logger()
 output = script.get_output()
+
+checked_families = 0
+count_errors = 0
+count_warnings = 0
+cancelled = False
+
+def update_output_title():
+    output.set_title("Family QuickCheck (Err:{} Warn:{})"
+                     .format(count_errors, count_warnings))
+
 
 all_families = DB.FilteredElementCollector(revit.doc)\
                  .OfClass(DB.Family)\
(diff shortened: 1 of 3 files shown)


extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/ptools.stack/Family.pulldown/Family QuickCheck.pushbutton/FamilyQuickcheck_script.py
```diff
@@ -6,37 +6,66 @@
...
 total_count = len(editable_families)
-checked_families = 0
 
 output.reset_progress()
 
-for fam in editable_families:
-    fam_name = revit.query.get_name(fam)
-    logger.debug("attempt to open family: %s", fam_name)
+with FailureSwallowerContext(force=True) as swallower:
     try:
```
melnikovalex, line 38: I find it is important to wrap it into try just bellow `with`. To be sure that the event will be unregistered in the end

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/ptools.stack/Family.pulldown/Save Families.pushbutton/SaveFamilies_script.py
```diff
@@ -4,41 +4,83 @@
...
+count_errors = 0
+count_warnings = 0
+
+
 for family in revit.query.get_families(revit.doc, only_editable=True):
     if family.FamilyCategory:
         family_dict[
             "%s: %s" % (family.FamilyCategory.Name, family.Name)
             ] = family
+if not family_dict:
```
melnikovalex, line 24: I recovered `script.exit()` here and below. In other case indentation would be to long. Sorry

pyrevitlib/pyrevit/revit/__init__.py
```diff
@@ -93,5 +93,72 @@ def get_project_info():
...
+    """Suppresses warnings during script execution
+
+    >>> with FailureSwallowerContext(force=True) as swallower:
+    >>>     for fam in families:
+    >>>         revit.doc.EditFamily(fam)
+    >>>         if swallower.get_swallowed():
+    >>>             logger.warn("Warnings swallowed")
+    """
+    # static, because the event cannot access non-static class variables
+    failures_swallowed = []
```
melnikovalex, line 110: Here is the most unstable part.

I couldn't find a way to store suppressed warnings as class properties and to reach them from `FailuresProcessing ` Event. 

For the same reason all the methods which should be triggered from the event are static.

Maybe you can propose a better solution

eirannejad:
@melnikovalex Thank you so much for the work on this. Great idea to have a failure swallower context manager. I merged this into its own branch, review your changes and manually transferred to the develop branch to be able to make a few adjustments and avoid merging old and new.

The commit is here: https://github.com/eirannejad/pyRevit/commit/eedf205f2950fdba5022c20ee005ef86a189eb54

Pull the develop branch and test out the changes and the Family Quick Check tool on your models. I modified your swallower to use the existing one that I had in `pyrevit.revit.db.transaction.py` so there is only one swallower in the backend and avoid duplication.

Let's get it tested and fine tune the results.

PR #754, opened 2019-11-14
ListElementsofSelectedLevel error

alexdaversa:
Ehsan,

I added the following line to the top of the ListElementsofSelectedLevel tool script to fix an error when trying to run the script:

`output = script.get_output()`


eirannejad:
Merged into develop. Thanks @alexdaversa

PR #757, opened 2019-11-20
fixed 'get_elements_by_categories' in forms

alexdaversa:
fixed reference in forms __init__ to 'get_elements_by_category' to the correct revit db query 'get_elements_by_categories'


eirannejad:
Thanks @alexdaversa. Merged into develop

PR #763, opened 2019-12-03
Copy/Paste state improvements

melnikovalex:
Hey Ehsan.

I was trying to extend the functionality of Copy/Paste Viewport position in PR #722 . The main aim was to be able to align viewports not only by the viewport center but also by the project base point. And add support of Sections as well as any other view type.

I stuck with PR #722 then, because I understood that it would be good to rework all the structure of Copy/Paste tools. That is what I tried to achieve here.

The first reason to rebuild it and to make it modular is that the functionality was split into two scripts - Copy and Paste. 

Therefore I left in these two .py scripts only 'UI-logic' and moved the main logic into `...stack\lib\copy_paste_state_actions.py`. In this case, I am using similar way as you have in Wipe-actions.

**Entities** 

There are several entities I am announcing to reduce repeating and make the content of each action simpler:

- `CopyPasteStateAction` - Base class for creating Copy/Paste State Actions.

Usage sample:
```
@copy_paste_action
class ViewScale(CopyPasteStateAction):
    name = 'View Scale'

def copy(self, datafile):
    view_scale = revit.active_view.Scale
    datafile.dump(view_scale)

def paste(self, datafile):
    view_scale_saved = datafile.load()
    revit.active_view.Scale = view_scale_saved

@staticmethod
def validate_context():
    if isinstance(revit.active_view, DB.TableView):
        return "Geometrical view must be active. Not a schedule."
```
- `DataFile` - Wraps pickle module and pyrevit datafile for using with `with`

Usage:
```
with DataFile('something', 'w') as datafile:
    view_name = revit.active_view.Name
    datafile.dump(view_name)
with DataFile('something', 'r') as datafile:
    view_name_saved = datafile.load()
```

`dump` and `load` also includes serialization and deserialization of certain Revit types. Some of those types were already there (e.g. `picklable_list`). I extracted it into `pyrevitlib.revit.serialization` because it is being used not only in Copy/Paste but also in Sync views. Please check how correct it is using it like that.

- Paste Tool now remembers which action was copied recently and you don't have to select it each time. But with Shift-Click you can select it manually.
- `validate_context` method allows showing only actions which are applicable in current context

**Tools**

Here I will describe improvements in the tools:

- Copy view zoom/pan state / SyncViews

  - I've changed both tools to have the same logic and storage format, implemented new `pyrevitlib.revit.serialization`
  - Added support of Section, Drafgin views.
  - Added support of 3D view orientation sync and copying.
  - Check view direction for Section before pasting

- Viewport placement

  - I completely rewrite the algorithm. Now it doesn't store 'transformation matrix` it uses either viewport center or does alignment using 'zero-cropbox' (for alignment by basepoint)

  - Support Section, Elevation, 3D views, Schedules (it should work with all placeable views now)

  - Support Sheets with different TitleBlock locations - the alignment is done according to TitleBlock (if it exists and in count no more than one)

   - Allow choosing ProjectBasePoint as an origin (instead of Cropbox)

   - Allow choosing cropbox alignment when Paste (top left, bottom left, etc.)

   - Flexible selection: can use either active view or selected viewport(s)

   - Support Pasting to multiple selected viewports at once (relevant for all Copy/Paste modes)

   - Bonus: running Paste with Shift-Click allows to isolate transformation to a certain axis (X or Y). So the tool can be used also for alignment many viewports on one sheet.

-  Crop Region

  -  Support Pasting to multiple selected views

- Filter Overrides

  - New action: copies certain VG-Filters overrides either to selected views or to the templates from the list

Please have a look.  Looking forward to your comments. I hope it won't take long to merge this.


eirannejad:
Merged into `pr/763` for review

eirannejad:
Merged into develop. @melnikovalex Thank you so much for the hard work on this

PR #769, opened 2019-12-11
"Remove Unsheeted Views" wipe action

melnikovalex:
Hey Ehsan! 

I've added a new action. It checks if the view id exists among Viewports (on sheets) and Viewers (if it has Callout which references the base view)

I've also tried to optimize views collectors with "lazy loading". 

Please have a look.


eirannejad:
Merged into `pr/769` for review

eirannejad:
Merged into `develop` Thanks a lot for the PR!

PR #806, opened 2020-01-28
change textbox to passwordbox in extension manager

Dre-Tas:
in the extension manager in the pyRevit GUI the passwords were shown as plain text. By changing the xaml TextBox element to a PasswordBox element you can hide the password showing it as dots.


eirannejad:
Beautiful. Thanks so much @Dre-Tas for being part of this project :D

PR #823, opened 2020-02-07
Fix pick_file, pick_excel_file

htlcnn:
Currently, `pick_file` has empty default `file_ext`, so when we `pick_file()`, filter would be `*.` (not `*.*` as it should be) meaning no file is available for selection.
Similarly, `pick_excel_file` should have its filter fixed to only `xlsx` and `xls`.
This PR fix that behaviour.
Please have a look @eirannejad 
Thanks!


eirannejad:
Merged to `pr/823` for review

eirannejad:
Merged into `develop`. Great change. Thanks a lot!

PR #829, opened 2020-02-17
Manage Packages - sorting revisions by Sequence number

davidvadkerti:
Hi,
in our company we use pyRevit. We started to use Manage Packages feature to manage revisions. I noted that if you rearange Revisions (Move Up & Move Down) they won't be sorted as they are in Revit. It seems they are sorted by ID now.

I think sorting revisions by Sequence number could fix this issue. I tried to fix it. After some testing on our larger projects it seems it works.

Another change I would propose is replace original Description in schedule header by Sequence number. I know that every company has different workflow. We f.e. write few words describing the changes in the revisions in the description. This makes project with few dozens of revisions very untidy and you need to scroll a lot. Originally I tried to rotate text of the heading by 90°. But I think showing Sequence number might be maybe better option. Hopefully you will find your revision faster.

**Revisions not sorted properly**

**Exact same file - the order of the revisions should be the same but it isn't**
Capture

**Suggested change - revisions are sorted by Sequence Number and Sequence Number is in the schedulle heading.**
460758066-Capture
I also added option to view just alphanumeric revisions as we f.e. use numeric revisions for revision clouds which we don't want to publish or track. I commented these lines as it is most likely not straightforward and not useful for anybody.

The reason of doing this was too long time of processing all the data in larger project. For examle in project with over 100 revisions and 800 sheets it took a few minutes to open the Manage Packages window. When I played with it I noticed that maybe number of commit points are crucial and not number of revisions or sheets. I will try to add some kind of filtering before window opens to cut the time for larger projects.

Sorry for long message and thank you for developing pyRevit :)
David

### DIFF
--- a/extensions/pyRevitTags.extension/lib/pkgmgr.py
+++ b/extensions/pyRevitTags.extension/lib/pkgmgr.py
@@ -176,13 +176,22 @@ def get_commit_points():
                         idx=idx,
                         name=docpkg.pkg_name))
     # grab revisions
+    # added x.SequenceNumber instead of x.Description
     docrevs = revit.query.get_revisions()
+    docrevsFiltered = []
+    for i, x in enumerate(docrevs):
+        # filter just Alphanumeric revisions
+        # if str(x.NumberType) == 'Alphanumeric':
+        #     docrevsFiltered.append(x)
+        docrevsFiltered.append(x)
+    sortedDocrevsFiltered = sorted(docrevsFiltered, key=lambda x: x.SequenceNumber)
+
     commit_points.extend([
         CommitPoint(cptype=CommitPointTypes.Revision,
                     target=x.Id.IntegerValue,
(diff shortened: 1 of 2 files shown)


eirannejad:
Merged into `pr/829` to review

eirannejad:
Accidentally merged into `develop`. Made a bit of a mess here

eirannejad:
Okay made some adjustments:
- The revisions are now sorted by sequence number and added a note to the window to make it clear

Screen Shot 2020-03-12 at 15 10 32

- Revisions columns now show the full revision description on the tooltip for quick checking

Screen Shot 2020-03-12 at 15 10 27

- Package columns now show the bound Project Parameter name on the tooltip for quick checking

Screen Shot 2020-03-12 at 15 10 21

davidvadkerti:
Thanks, I've just tried your adjustments it works very well :)

PR #893, opened 2020-05-10
Sync Views bugfix, Copy State improvements

melnikovalex:
Hey Ehsan

Thank you for your last updates and for fixing my code. I haven't had a chance to use all of it in a new release, but run into one bug.

It looks like we messed up Sync View (also in Copy State) a bit. It stopped working at all in the new release.

I did the fixes (da06d5c, d4cf8e5) and added a couple of improvements:

  * e745bc1, 29bfe63 Paste State script - show only actions with data

> Added a new method into `pyrevit.script` which returns true if datafile exists. Can be moved out of there if it's too narrow case.

  * edfff54 Copy State - allow to copy between projects

> I can name several use cases when it's necessary: to navigate in two linked model (e.g. structural and architectural, it should also work in two Revit sessions), to compare visual differences with an older version of a model. 

> It won't work with Filters and V/G Settings, unfortunately, so there are a couple of extra lines to say which method only for `this_project`.

> I was wondering if it is possible to do the same for 'Sync Views'. Doing it directly could be annoying if you switching between not connected projects. I was thinking about enabling it with Shift Click but couldn't find a good solution for that. 

  * 427de7c Copy Filter Overrides - support active view, fixes

> When I code this part before, the idea was to be able to copy from active_view and paste to active view. So I recovered this code. If no views selected, it gets or sets data for the active one.

If you're seeing all these commits from 2019 in PR, sorry for that. One day I'll learn how to do it properly 😄 

Thanks in advance!

Alex


eirannejad:
Hey @melnikovalex Thank you so much for improving the codebase and submitting your changes, but, massive PRs like this make it very very hard for me to review the changes. I basically have to go through every commit, determine what are the changes, make adjustments and test, and move on to the next. A single PR, needs to focus on a single change to it can be easily reviewed and tested.

This means that reviewing a PR like this, most likely, is not a single-day job and gonna take a lot of time. Also, it makes it very hard for other contributors to help in reviewing the PR since it is all in one chunk. If this is broken down into 5-6 individual PRs, they can be reviewed independently by multiple developers.

Would you mind breaking this into PRs organized based on the type of improvement or new feature/tool they add?

melnikovalex:
> Would you mind breaking this into PRs organized based on the type of improvement or new feature/tool they add?

Sorry, missed your message. Here is it: #927

PR #894, opened 2020-05-10
"Select Similar in a view" tool - use quick filter, allow to select in whole doc

melnikovalex:
Hey Ehsan,

Several times I found myself stuck with a question: how to select all walls in a model. Go into a 3D view? I had the Ideate Explorer Addin for that, but it also looked like overkill.

So I was trying to find something similar among pyRevit buttons. It required just a little change :)

Here is my small improvement. Please have a look.

- now it uses quick filters
- it allows to search in the whole document (I did that with Shift-Click)

Hope I didn't mess anything. Thanks for merging and have a nice day!

Alex


eirannejad:
This is great. I'll review and merge this today. Great PR. Thanks @melnikovalex

PR #927, opened 2020-06-22
Sync Views bugfix, Copy State improvements

melnikovalex:
Hey Ehsan,

Sorry, missed your message in the last PR. Here are the same changes as in #893, with no outdated commits + one more bugfix. I've seen you've already made some changes in these files. If so, please have a look at the last commit:

* e348ffa - Fix CopyState - ViewZoomPanState
 
_Here are the old changes, from previous PR:_

So it looks like we messed up Sync View (also in Copy State) a bit. It stopped working at all in the new release.

I did the fixes (bf7a39e , cced091) and added a couple of improvements:

  * e745bc1, d9dbb9e Paste State script - show only actions with data

> Added a new method into `pyrevit.script` which returns true if datafile exists. Can be moved out of there if it's too narrow case.

  * f35ca4a Copy State - allow to copy between projects

> I can name several use cases when it's necessary: to navigate in two linked model (e.g. structural and architectural, it should also work in two Revit sessions), to compare visual differences with an older version of a model. 

> It won't work with Filters and V/G Settings, unfortunately, so there are a couple of extra lines to say which method only for `this_project`.

> I was wondering if it is possible to do the same for 'Sync Views'. Doing it directly could be annoying if you switching between not connected projects. I was thinking about enabling it with Shift Click but couldn't find a good solution for that. 

  * a20cc71 Copy Filter Overrides - support active view, fixes

> When I code this part before, the idea was to be able to copy from active_view and paste to active view. So I recovered this code. If no views selected, it gets or sets data for the active one.

Thanks,
Alex

### DIFF
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Selection.panel/memo.stack/Paste State.pushbutton/script.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Selection.panel/memo.stack/Paste State.pushbutton/script.py
@@ -6,7 +6,7 @@ Shift-Click:
 Show additional options
 """
 from pyrevit import PyRevitException
-from pyrevit import forms
+from pyrevit import forms, script
 
 import copypastestate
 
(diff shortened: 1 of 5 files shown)


eirannejad:
Thank you so much @melnikovalex I know preparing these meticulous PRs are hard and I really appreciate you supporting this project 🙏

Issue #1015, opened 2020-09-29
Load all families from folder tree


DuncanLithgow:
**Is your feature request related to a problem? Please describe.**
When testing families that would normally be in a content management system it's useful to be able to load many families at once. A tools for this would be great.

**Describe the solution you'd like**
There is a 'Load more types' Family tools. I'd love a 'load more families' tool which accepts a folder as input and then asks if all families should be loaded recursively. Trying to filter backups out would of course be a good idea, other wise a warning is fine.

**Describe alternatives you've considered**
Loading them one at a time ... boring

**Additional context**
Suggestions:
Name: Load Multiple Families
Description: Loads multiple Families from a folder structure, while trying to avoid backups
Options: yes/no: load recursively. Radio buttons: load only family or load all types from Type Catalogue

An extension of functionality would be a dialogue like the purge dialogue which allows for selecting individual types to be loaded.

hikmet-kibar:
@DuncanLithgow 
Liked the idea, made the tool -> [LoadFamilies.pushbutton](https://github.com/hikmet-kibar/LoadFamilies.pushbutton)
Since I don´t have a large set of tools, I did not create a whole extension. You can [create you own extension ](https://www.notion.so/Create-Your-First-Command-2509b43e28bd498fba937f5c1be7f485), if you don´t already have a custom one, and add this tool.
I could only test it for myself - Let me know how it goes for you.

@eirannejad 
Many parts are based on your `Load More Types` tool. Credit can be found in the docstrings of the classes.

eirannejad:
@hikmet-kibar This is great. I'll grab it from your repo and will add to the toolbar

david939david:
@hikmet-kibar Thank you for the link to "create your first command", it helped me a lot with understanding what is going on ;)

PR #1041, opened 2020-10-25
added Load Families.pushbutton

hikmet-kibar:
New bundle to load families from a directory, recursively.
Related to issue #1015 .

**Workflow**
1. Select a directory.
2. Select Families (*.rfa files) from the directory.
3. Select loading option (load all Types OR select Type(s) for each Family)
(4. Select Family Type(s) for each Family)

**Notes on bundle**
- Tested on Revit 2019, 2020 with up to 450 families, loading all types.
- Reports at the end about families that were already loaded _(if family name exists in project)_.
- If user decides on selecting types for each family, single type families will be loaded without prompt.
- Excludes backup files ([xxx.0001.rfa](https://knowledge.autodesk.com/support/revit-products/getting-started/caas/CloudHelp/cloudhelp/2019/ENU/Revit-GetStarted/files/GUID-E79353E9-CC57-4ECE-93B8-5D12D9B3B135-htm.html))
- Has progress bar with cancel button and count.

**Notes on the code**
- `pathlib` [module ](https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob) to find files recursively by globbing
  - pathlib is part of the standard lib of python since v3.4. Can be replaced with [os.walk ](https://docs.python.org/3/library/os.html#os.walk)or [os.path.walk](https://docs.python.org/2/library/os.path.html#os.path.walk), depending on python version.
- [itertools.ifilterfalse](https://docs.python.org/2/library/itertools.html#itertools.ifilterfalse) to filter out backup files. Can be replaced by `re.match`.


hikmet-kibar:
I moved the pushbutton into a new pulldown with 'Load More Types'. 
The pulldown is missing an icon. Any suggestions are welcome and i´ll add it in a commit with other changes after more reviews.

eirannejad:
Uh sweet!

eirannejad:
Merged for review

eirannejad:
Merged into develop and will be published with v4.8.4

PR #1076, opened 2020-11-26
Parameter framework modification

thumDer:
I've modified the `select_parameters` function and the underlying `ParamDef()` a little:

- Now the IsReadOnly filter is optional instead of  hardcoded, but the default behaviour stays the same (not showing ReadOnly). We have a lot of usecases where we need to get the ReadOnly parameters as well.
- I've modified the Parameter `item_template` to show if a parameter is ReadOnly, and also added a few extra info about the parameter itself into a ToolTip, to help identify the parameters.
- It follows that I had to modify the `ParamDef()` named tuple aswell, to contain the readonly attribute, and the parameter definition itself. The reason for the last one is that it is a much better way to get an elements parameter by their definition (`parameter = Element.get_Parameter(Definition)`) instead of their name with `LookupParameter(string)`, because in actual projects parameter name duplications happen. A lot. This is also a reason for the added ToolTip

I've kept the order of the attributes for the `ParamDef()` to keep it backward compatible for commands that get the attributes from the tuple by their index instead of their name.


eirannejad:
This is a great PR. Thank you! @thumDer

PR #1117, opened 2021-01-07
Modification to Import/Export Family Configs

msmith320:
The original import and export family configs buttons did not account for shared parameters or parameters whose values were of type ElementId. The following scripts use the GUID of shared parameters and a specific string to keep track of the shared parameters and the Load Classification Type. The export button saves a txt file in addition to the yaml file with the same name and in the same directory. For import, make sure both the txt and yaml file are in the same directory.


msmith320:
Please let me know if I need to make any changes.

eirannejad:
Merged into `pr/1117` for review and will be merged into develop after

msmith320:
@eirannejad Thanks for looking at the scripts. I commented a few adjustments in your commit. There was a naming issue with `DB.Electrical.ElectricalLoadClassification` and an issue with a variable being defined in an if block.

eirannejad:
Ok great. This is already merged so would you mind sending a new PR to `develop` with your changes?

PR #1121, opened 2021-01-12
Develop full soft tab colors

modulusArch:
Have played with the settings and finding that using pastel colors and a complete tab infill assist in stronger project and family identification. A soft blue or soft green might be a good addition, howver started with soft yellow, pink, and tan colors.
image


eirannejad:
Added this to list of items to be fixed when upgrading the colorizer

PR #1156, opened 2021-02-04
Added UTF-8 encoding

monsieurhannes:
I added UTF-8 encoding with `# coding=utf-8` and commented some tips how to add prefix/suffix types.
I didn't find another way to add e.g. umlauts.


eirannejad:
Thanks a lot!

Issue #1167, opened 2021-02-12
Print Sheets


Papala18:
Using the Print Sheets command I am printing by sheet list.  

- The sheet list has a filter by issue date
- The printer is Bluebeam PDF
- The Page settings are a company standard sheet, grayscale
- I checked the combine into one PDF box

It all prints out as expected.  The problem is in the Project Browser.  The sheets that are printed thru the Print Sheets command have a character prefix added (an arrow pointing right) to the sheet number.  This character doesn't print but the printed sheets sent to the bottom of the Project Browser sheet list.  I have to rename each sheet that was printed to the original sheet number to get them to go back to their original sheet order.

monsieurhannes:
I have the same problem. It got so bad that the sheet number increased to more than 200 (which might be the limit for Revit). After a while the sheet number (gradually) vanished from the Project Browser making it impossible to open the sheet!
![screenshot_1](https://user-images.githubusercontent.com/20651440/108859448-84c2f400-75ed-11eb-8ca1-05919ca52998.png)

Seems like a serious bug as it's not visible until the sheet vanishes. 
I wrote a Dynamo-Script that cleans the sheet numbers. As I didn't know if it's only prefix or even suffix that get added just ran it against all characters + some special characters from my region. It might help some, adapt the special characters to your needs.
![remove_special_character](https://user-images.githubusercontent.com/20651440/108860768-cf913b80-75ee-11eb-8261-859861a60597.png)

```python
# Load the Python Standard and DesignScript Libraries
import sys
import clr
clr.AddReference('ProtoGeometry')
from Autodesk.DesignScript.Geometry import *

import string

# The inputs to this node will be stored as a list in the IN variables.
dataEnteringNode = IN

# Place your code below this line
list_in = IN[0]
list_out = []

def listAlphabet():
  temp = list(string.ascii_lowercase) + list(string.ascii_uppercase)
  special = ["ö","Ö","ä","Ä","å","Å","-","_"," "]
  for i in range(10):
  	special.append(str(i))
  return temp+special
  
chars = listAlphabet()

for sheet_number in list_in:
	sheet_number_out = ""
	for character in sheet_number:
		if character in chars:
			sheet_number_out = sheet_number_out + character
		else:
			pass
	list_out.append(sheet_number_out)
	
# Assign your output to the OUT variable.
OUT = list_out
```

eirannejad:
[...]

PR #1179, opened 2021-02-28
Cleaning NPC before reapplying them

monsieurhannes:
This should prevent prefixes from previously failed prints being reapplied to the sheets. Associated to this issue #1167 

Another possible solution would be to just wipe NCP in the end, instead of saving all the sheet numbers and reapplying them? That saves two steps.


eirannejad:
This is a great idea. Thank you!

eirannejad:
Merged for review

Issue #1199, opened 2021-03-24
Duplicate Selected Views As Dependents


gsruff:
Drawing Set->Views->Duplicate Selected Views does not currently offer an option to duplicate views as dependents. I believe this would be a useful addition.

Here is my attempt at implementing this:

```
"""Batch duplicates the selected views with detailing, without detailing, or as dependents."""

from pyrevit import revit, DB
from pyrevit import forms
from pyrevit import script


logger = script.get_logger()


def duplicableview(view):
    return view.CanViewBeDuplicated(DB.ViewDuplicateOption.Duplicate)


def duplicate_views(viewlist, dupop):
    dup_view_ids = []
    with revit.Transaction('Duplicate selected views'):
        for el in viewlist:
            if dupop == 'with_detailing':
                dupop = DB.ViewDuplicateOption.WithDetailing
            elif dupop == 'without_detailing':
                dupop = DB.ViewDuplicateOption.Duplicate
            else:
                dupop = DB.ViewDuplicateOption.AsDependent

            try:
                dup_view_ids.append(el.Duplicate(dupop))
            except Exception as duplerr:
                logger.error('Error duplicating view "{}" | {}'
                             .format(revit.query.get_name(el), duplerr))
        if dup_view_ids:
            revit.doc.Regenerate()
    return dup_view_ids

selected_views = forms.select_views(filterfunc=duplicableview,
                                    use_selection=True)

if selected_views:
    selected_option = \
        forms.CommandSwitchWindow.show(
            ['With Detailing',
             'Without Detailing',
             'As Dependent'],
            message='Select duplication option:'
            )

    if selected_option:
        dupop = 'with_detailing'
        if selected_option == 'Without Detailing':
            dupop = 'without_detailing'
        else:
            dupop = 'as_dependent'
        dup_view_ids = duplicate_views(
            selected_views, dupop)
        if dup_view_ids:
            revit.get_selection().set_to(dup_view_ids)

```

gsruff:
Thanks guys!

PR #1202, opened 2021-03-26
Update Duplicate Views Dependent

chuongmep:
Dear Ehsan, 
Im Update Duplicate Views Dependent with request from user #1199 
Test Demo:
DuplicateViewDemo


eirannejad:
Thank you so much @chuongmep ! This is great

PR #1221, opened 2021-04-19
allowdialogs fix

thumDer:
There was a tiny line missing, now its perfect!


eirannejad:
@thumDer Tiny but sooooooo important 😆 Sorry for the mess and thanks for finding the bug :D

PR #1242, opened 2021-05-16
Fix config reading for switches in CommandSwitchWindow

deddolo:
typo in conditional statement


eirannejad:
Thank you! Great fix.

PR #1247, opened 2021-05-21
Forms getvaluewindow slider

alexdaversa:
Hi Ehsan,

This pull request is to add a number slider option to the `GetValueWindow` class in the forms module. It also adds a function `ask_for_number_slider` to make this available to extensions.

I have picked defaults that work for my use case for the following slider properties:
`TickPlacement="None" TickFrequency="5" IsSnapToTickEnabled="True"` 
I have only exposed slider min/max to the `ask_for_number_slider` function, but could add more...I just didn't want it to be an overwhelming amount of options.

I am not as familiar with .NET and WPF--should I include a check to make sure that min is always <= to max?


eirannejad:
Thank you so much 🙏

PR #1256, opened 2021-05-28
Hotfix/telemetry username

Dre-Tas:
I've done a bit of research on why the host_user field returns null and, although I haven't found an actual reason, many forums suggest not to use Environment.UserName as it is unstable, but instead use WindowsIdentity. So that's what I've done.

References here:
https://social.msdn.microsoft.com/Forums/vstudio/en-US/62867979-d5cf-4b40-b966-96ea5c129ee6/environementusername-is-returning-nullempty-string-in-windows-7?forum=vbgeneral
https://www.reddit.com/r/Unity3D/comments/e5yg2l/c_environmentusername_returning_as_system_rather/
https://social.msdn.microsoft.com/Forums/vstudio/en-US/4e0a955c-6740-4af7-9216-653c05ef98be/myusername-is-returning-a-null-empty-string?forum=vbgeneral


eirannejad:
Thank you!

PR #1276, opened 2021-06-15
Titleblock on sheets

alexdaversa:
Ehsan,

This PR adds a shiftclick mode to the List/Select TitleBlocks on Sheets tool to directly select the titleblocks on sheets selected, bypassing the output window. This is useful for situations where the titleblock in use on sheets is already known, and a user wants to edit instance parameters of the titleblock in bulk.

Thanks,
Alex


eirannejad:
Thank you so much 🙏

PR #1464, opened 2022-01-30
Dev/issues/1442

alexdaversa:
I spent awhile, growing more surprised by the minute, searching the Revit API for a way to clear the current selection. This workaround does work, but seems particularly hacky.


alexdaversa:
I think i based this on the wrong branch, sorry

PR #1465, opened 2022-01-30
Dev/issues/1442

alexdaversa:
trying this again...

I spent awhile, growing more surprised by the minute, searching the Revit API for a way to clear the current selection. This workaround does work, but seems particularly hacky.


extensions/pyRevitTools.extension/pyRevit.tab/Drawing Set.panel/Print Sheets.pushbutton/script.py
```diff
@@ -1397,6 +1397,7 @@ def cleanup_sheetnumbers(doc):
 
 # verify model is printable
 forms.check_modeldoc(exitscript=True)
+selection = revit.uidoc.Selection.SetElementIds(List[DB.ElementId]([revit.DB.ElementId(2147483647)]))
```
eirannejad, line 1400: @alexdaversa I'm going to change this to use `DB.ElementId.InvalidElementId` and will add this to `pyrevit.revit.selection` as helper method.

RE: https://forums.autodesk.com/t5/revit-api-forum/how-to-clear-a-selection/td-p/9926146

eirannejad: https://github.com/eirannejad/pyRevit/commit/704f39fb8437123829b7c0d798c55bfd80c1bef6

PR #1508, opened 2022-04-06
Adds preprocessor directives for Revit 2023 to avoid the depracated T…

dagz214:
…aggedLocalElementId property for IndependentTag

Tested and working with Revit2023 + Revit2021


BlackMac0509:
I basically copied the complete dagz214 code, and pasted it below the original code in ScriptConsoleutils.cs, then deleted the original.

The only problem encountered was in saving. Had to modify the permissions of the original file to allow it to be overwritten.

Issue #1438, opened 2021-12-31
Make pyrevit module compatible with python 3


eirannejad:

sanzoghenzo:
Given that pythonnet v3 milestone due date has not been disclosed yet, would you accept a PR to fix the current incompatibility?

I'm not sure if I can solve everything, but I could try it.

But first, a quick question: how can I load a git clone of pyrevit(lib)?

Forgive me if there's already an answer somewhere, but I can't find any info.
The first place I tried was CONTRIBUTING.md, but it's empty..

emilaz:
The pythonnet 3.0 should be released any day now, with `v3.0.0-rc4` already out (https://github.com/pythonnet/pythonnet/releases). Maybe we can start switching to pythonnet v3, as the remaining changes to v3 will be miniscule? The faster we move away from the badly maintained security risk that is IronPython2.7 the better :)

PR #1545, opened 2022-06-15
fix: add cpython compatibility

sanzoghenzo:
This is my first attempt at closing #1438

It's not at all a comprehensive fix for now, I only tried a direct import of the following:

```py
from pyrevit import EXEC_PARAMS
from pyrevit import revit
from pyrevit import script
from pyrevit.revit import Transaction
```

There's an ugly hack to get the latest cpython engine version: I couldn't get the integer from the `Version` property, so I used `str()` to compare them.


sanzoghenzo:
I see that the CI can't find my branch, should I name it develop (and create a PR against it)?

PR #1556, opened 2022-06-30
Tool to delete Arrowheads Types - Bazooka Style

jmcouffin:


jmcouffin:
image
Both in French and English except for the form

PR #1634, opened 2022-10-05
Attempt to fix tags manager

jmcouffin:
Not my cleanest but should solve a couple of issues such as https://github.com/eirannejad/pyRevit/issues/1364 and https://github.com/eirannejad/pyRevit/issues/1632

Added "you need to create a 'doctag' and apply it to the document categories before using the tag tools" to bundle files
It took me a bit to figure out why I couldn't use the tools at first

### DIFF
--- a/extensions/pyRevitTags.extension/lib/tagsmgr.py
+++ b/extensions/pyRevitTags.extension/lib/tagsmgr.py
@@ -555,8 +555,11 @@ def apply_tags(element, tags, config=None):
                                                 space=False,
                                                 spare=False)
         if esystems:
-            for esys in esystems:
-                apply_tags(esys, tags, config=config)
+            try:
+                for esys in esystems:
+                    apply_tags(esys, tags, config=config)
+            except Exception as apply_err:
+                raise(apply_err)
 
     if config.append:
         el_tags = extract_tags(element)
--- a/pyrevitlib/pyrevit/revit/db/query.py
+++ b/pyrevitlib/pyrevit/revit/db/query.py
@@ -970,10 +970,17 @@ def get_connected_circuits(element, spare=False, space=False):
         circuit_types.append(DB.Electrical.CircuitType.Spare)
     if space:
         circuit_types.append(DB.Electrical.CircuitType.Space)
-
-    if element.MEPModel and element.MEPModel.ElectricalSystems:
-        return [x for x in element.MEPModel.ElectricalSystems
-                if x.CircuitType in circuit_types]
+    if circuit_types !=[DB.Electrical.CircuitType.Circuit]:
+        if HOST_APP.is_newer_than(2021, or_equal=True): # deprecation of ElectricalSystems in 2021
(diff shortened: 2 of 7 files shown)


extensions/pyRevitTags.extension/lib/tagsmgr.py
```diff
@@ -555,8 +555,11 @@ def apply_tags(element, tags, config=None):
                                                 space=False,
                                                 spare=False)
         if esystems:
-            for esys in esystems:
-                apply_tags(esys, tags, config=config)
+            try:
+                for esys in esystems:
+                    apply_tags(esys, tags, config=config)
+            except Exception as apply_err:
+                raise(apply_err)
```
eirannejad, line 562: what does this change do? Seems like it captures and re-raises the exception without any processing. So the `try except` block doesn't really do anything

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -970,10 +970,17 @@ def get_connected_circuits(element, spare=False, space=False):
...
     if space:
         circuit_types.append(DB.Electrical.CircuitType.Space)
-
-    if element.MEPModel and element.MEPModel.ElectricalSystems:
-        return [x for x in element.MEPModel.ElectricalSystems
-                if x.CircuitType in circuit_types]
+    if circuit_types !=[DB.Electrical.CircuitType.Circuit]:
+        if HOST_APP.is_newer_than(2021, or_equal=True): # deprecation of ElectricalSystems in 2021
+            if element.MEPModel and element.MEPModel.GetElectricalSystems:
+                return [x for x in element.MEPModel.GetElectricalSystems
```
eirannejad, line 976: `GetElectricalSystems` is a method. https://www.revitapidocs.com/2023/ef8e7119-9fa4-3024-d43d-591759b6098b.htm

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -970,10 +970,17 @@ def get_connected_circuits(element, spare=False, space=False):
         circuit_types.append(DB.Electrical.CircuitType.Spare)
     if space:
         circuit_types.append(DB.Electrical.CircuitType.Space)
-
-    if element.MEPModel and element.MEPModel.ElectricalSystems:
-        return [x for x in element.MEPModel.ElectricalSystems
-                if x.CircuitType in circuit_types]
+    if circuit_types !=[DB.Electrical.CircuitType.Circuit]:
```
eirannejad, line 973: What does this do?

pyrevitlib/pyrevit/revit/db/query.py
jmcouffin: rookie mistake

extensions/pyRevitTags.extension/lib/tagsmgr.py
jmcouffin: you are right

Issue #1555, opened 2022-06-29
Sheet reorder not working


sakhyr:
The tool Sheet Reorder is not showing any sheet on my project to reorder, and parameters is empty.
Is there any solution for that?

![empty reorder](https://user-images.githubusercontent.com/108416046/176466041-bc3de996-d30a-4d8d-9beb-39b1eb3da457.jpg)
![empty parameters](https://user-images.githubusercontent.com/108416046/176466068-f409625f-721e-4e93-ba95-b6cb5a39b26e.jpg)

Marcelo-Romero:
I think this issue happens because, most of the time, these parameters are created with the String type. Maybe if we just change "Integer" for "String" in line 101, the tool will work.

PR #1640, opened 2022-10-08
Fix ReOrder Sheets

Marcelo-Romero:
This change cancels the program if there are no "order" parameters of the "integer" type for the sheets. If it's cancelled, it gives the user a window message explaining why this happened. Issue #1555


eirannejad:
@Marcelo-Romero So one of the things to watch out for is the amount of change that is introduced in a single commit/PR. Take a look at the [red and green markers here](https://github.com/eirannejad/pyRevit/pull/1640/commits/c69e087979f5b95fe6b6708bfbbdb4c5a183b41d). It seems that a series of lines has been added to the start and everything else (event class definitions) are moved inside of the `else` block. This is generally not a good idea to have types nested in a block. It also creates a lot of (-) removed and (+) added lines in the commit that basically do not change the script. It makes reviewing the changes very hard.

Marcelo-Romero:
> @Marcelo-Romero So one of the things to watch out for is the amount of change that is introduced in a single commit/PR. Take a look at the [red and green markers here](https://github.com/eirannejad/pyRevit/pull/1640/commits/c69e087979f5b95fe6b6708bfbbdb4c5a183b41d). It seems that a series of lines has been added to the start and everything else (event class definitions) are moved inside of the `else` block. This is generally not a good idea to have types nested in a block. It also creates a lot of (-) removed and (+) added lines in the commit that basically do not change the script. It makes reviewing the changes very hard.

I see what you are saying. I just didn't know other way to catch a possible scenario and make a decision from that. Let me know if the idea of having the script giving this info to the user about what is going on is still good for now. If it is, I'm gonna look into how to write the fix without changing so many lines.

eirannejad:
@Marcelo-Romero Merged your PR but rewrote the logic to it is more pythonic and simpler :D 
See my changes here: https://github.com/eirannejad/pyRevit/commit/7b79ac375193f6e3c7be655a7f84ea04837c42d3

PR #1675, opened 2022-12-04
DB.SheduleField support

ejs-ejs:
Partial support for pickling DB.ScheduleField.

### DIFF
--- a/pyrevitlib/pyrevit/revit/db/pickling.py
+++ b/pyrevitlib/pyrevit/revit/db/pickling.py
@@ -150,6 +150,54 @@ class ViewType(EnumSerializable):
     api_types = DB.ViewType
 
 
+class Grid(Serializable):
+    api_types = DB.Grid
+
+    def __init__(self, gridline):
+        from rpw import doc
+
+        cView = doc.ActiveView
+        curves=gridline.GetCurvesInView(DB.DatumExtentType.ViewSpecific, cView)
+        cCurve = curves[0]
+        start_curve = gridline.GetLeader(DB.DatumEnds.End0, cView)
+        end_curve = gridline.GetLeader(DB.DatumEnds.End1, cView)
+
+        if start_curve:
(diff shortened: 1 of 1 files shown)


pyrevitlib/pyrevit/revit/db/pickling.py
```diff
@@ -150,6 +150,76 @@ class ViewType(EnumSerializable):
...
+        return self.name, self.heading, self.hidden, self.width.deserialize(), self.FO, self.FO_Accuracy.deserialize(), self.FO_sym, self.FO_usePlus
+      else:
+        return self.name, self.heading, self.hidden, self.width.deserialize()
+
+
+class Grid(Serializable):
+    api_types = DB.Grid
+
+    def __init__(self, gridline):
+        from rpw import doc
```
jmcouffin, line 179: Use 'doc = revit.doc' from the pyRevit revit module?

ejs-ejs: the Grid class is already in the master, so it may be skipped.

eirannejad:
This PR seems odd. `DB.ScheduleField` does not have any constructors. How are you `deserialize`ing the stored data into a `DB.ScheduleField`?

ejs-ejs:
> This PR seems odd. `DB.ScheduleField` does not have any constructors. How are you `deserialize`ing the stored data into a `DB.ScheduleField`?

The deserializing is needed for numeric fields, like 'width' AFAIR . What I [think | hope], the API is creating a new DB.ScheduleField and the properties can be applied to the existing DB.ScheduleField. I need to clean the kind-of-working code, the repo is  https://github.com/ejs-ejs/py-ejs.

jmcouffin:
@ejs-ejs did you get the chance to validate this change and test it?

jmcouffin:
I will leave it alive for an extra week then close the PR

PR #1692, opened 2022-12-30
Autosave

alexdaversa:
Hi Ehsan,

Proposing to add an Autosave tool via hooks. It comes preconfigured on install disabled (user opt-in) with a default interval of 15 minutes. Clicking the pushbutton toggles functionality on/off and shift-click allows to set interval between saves.

I developed this for the firm I work for, KSS Architects, and we have been using this tool as part of a custom extension for about a year now.

By default, this will not try to save unsaved projects, detached models, and families. The shared library drive at our office is write-protected so I have not had the need to check file paths against an exclusion list, but it could be added in the future.

For non-workshared files, this saves over the project file (and Revit creates the .0001.rvt backups). For workshared files, both fileserver and BIM360/ACC-based, it saves over the local only. This avoids sync conflicts. If the file is closed without syncing, Revit prompts the user to either keep their changes or discard and start from the last sync.


eirannejad:
This is incredible. Thank you!

jmcouffin:
@alexdaversa 
Last meetup, we discussed a bit the functionality. 
While definitely a great piece, we were a bit scared to bring that in as part of the default features of pyRevit.
We would love to keep this available, but maybe more like a specific extension that is not enabled by default.
If you agree, would you mind creating the extension as a separate public repo and just make a PR to add it to the list of publicly available extensions?

in here https://github.com/eirannejad/pyRevit/blob/develop-4/extensions/extensions.json

alexdaversa:
@jmcouffin I'd be happy to provide this feature as an extension although it will take me some time as I have become busy with other personal commitments.

I would also be interested to learn from the greater team what about this tool is causing hesitancy to bring into the base install. I am open to modifying the functionality of the tool based on this feedback to improve it.

jmcouffin:
@alexdaversa 

> hesitancy to bring into the base install

Mostly: 
- Save temporary / unwanted changes and unable to go back (transaction history gets cleared after a save)
- Save occurring during another opened/complexed transaction

> provide this feature as an extension

I could help with that if you just create a public repo and allow me to PR

jmcouffin:
@alexdaversa it is being used https://discourse.pyrevitlabs.io/t/threading-autosave/7757 🍾

alexdaversa:
@jmcouffin I am finally getting around to creating the extension! Apologies to all for the wait. I should have just opened the repo to let you PR against.

I understand the hesitancy to bring this into the main pyRevit repo, and wish Autodesk had built this feature into the main product without those limitations! Until then...

Do you recommend that this tool live as an extra button on the pyRevit tab, a button on the Addins tab, or as its own tab?

alexdaversa:
closing this as the tool was moved to its own repo https://github.com/alexdaversa/pyRevit-Autosave which can be enabled from extensions in the latest build.

PR #1742, opened 2023-02-27
Fix Tooltip videos

dvdgnz:
See https://github.com/eirannejad/pyRevit/issues/1265


jmcouffin:
Just checked and tried it. It works fine!
Thank you so much for that one @dvdgnz 
This has been a long standing issue.

PR #1743, opened 2023-02-28
New Tools: Set legend component orientation

jmcouffin:
This new tool allow to set in batch orientation of legend component. 
It works even if no view is properly setup in the family itself.

https://user-images.githubusercontent.com/7872003/221819860-beccbdde-1410-4389-bb36-d6c4921d50b4.mp4

_Note:_
In some rare cases, it does not apply the orientation.


jmcouffin:
inspired by [BIMMajster tools](https://apps.autodesk.com/en/Publisher/PublisherHomepage?ID=5YUFLD8EQEQE) 

[Legend Components Hidden Views](https://apps.autodesk.com/RVT/en/Detail/Index?id=5686429036533794385&appLang=en&os=Win64)

PR #1752, opened 2023-03-07
Select Elements from Group and Subgroups

nodatasheet:
There is a nice button named `Select Group Elements`. It only selects elements inside selected groups. But often we need to select  elements from the group and its sub-groups and their sub-sub-groups, etc

I propose adding another button `Select Group And Subgroup Elements` which selects all the deepest elements from the selected groups.


jmcouffin:
@nodatasheet 
Are you done editing? in which case I will merge the PR

nodatasheet:
> @nodatasheet Are you done editing? in which case I will try it then merge the PR

Yes, all good! Can test and merge.

jmcouffin:
@nodatasheet it works fine.
Good Job!
I just added the french translation for the tool

PR #1753, opened 2023-03-09
New Feature override 2D elements in current view

jmcouffin:


jmcouffin:
smart button to highlight or clear overrides of 2D elements
this is a smartbutton to show whether the command was launched. 

_Note_: It does not take into account active view switching

PR #1770, opened 2023-03-23
Delete AT-AT_16488_content.rfa

jmcouffin:


jmcouffin:
simplified the family. 
The old one was nicer but at the expense of family size

PR #1777, opened 2023-03-30
fix assembly cache

dosymep:
for issue
https://github.com/eirannejad/pyRevit/issues/1776


dosymep:
I can't create unit tests because pyrevit loses the cache after a reboot and thus several assemblies appear in memory when restarted

dosymep:
Maybe this changes needs to thorough check:
1. when assembly non change (in memory one image)
2. when assembly was changed (in memory several image)
3. when pyrevit was reloaded and assembly non change
4. when pyrevit was reloaded and assembly was changed
5. etc

eirannejad:
Merged into develop and will be released with next version very soon. Thanks @dosymep

PR #1780, opened 2023-04-02
Removed usage error in load_data example

perrylackowski:
Example had the de-serialization function of the CustomData object listed as a second argument, which didn't match the load_data function definition.


jmcouffin:
Thanks

Issue #1705, opened 2023-01-17
IPY340 Set Revision On Sheets rises IOException when revision numbering is per sheet


nodatasheet:
**Describe the bug**
With the build version "4.8.13:16d0938" clicking on `Set Revision On Sheets` button rises the `Autodesk.Revit.Exceptions.InvalidOperationException`:
```
This operation is not valid for when the revision numbering is per sheet.
```

<details>
<summary><b>Error traceback</b></summary>

```Python
IronPython Traceback:
Traceback (most recent call last):
 File "C:\Users\Konstantin\AppData\Roaming\pyRevit\pyRevit\develop\extensions\pyRevitTools.extension\pyRevit.tab\Drawing Set.panel\Revision.pulldown\Set Revision On Sheets.pushbutton\script.py", line 7, in <module>
 File "C:\Users\Konstantin\AppData\Roaming\pyRevit\pyRevit\develop\pyrevitlib\pyrevit\forms\__init__.py", line 1963, in select_revisions
 File "C:\Users\Konstantin\AppData\Roaming\pyRevit\pyRevit\develop\pyrevitlib\pyrevit\forms\__init__.py", line 653, in show
 File "C:\Users\Konstantin\AppData\Roaming\pyRevit\pyRevit\develop\pyrevitlib\pyrevit\forms\__init__.py", line 633, in __init__
 File "C:\Users\Konstantin\AppData\Roaming\pyRevit\pyRevit\develop\pyrevitlib\pyrevit\forms\__init__.py", line 877, in _setup
 File "C:\Users\Konstantin\AppData\Roaming\pyRevit\pyRevit\develop\pyrevitlib\pyrevit\forms\__init__.py", line 1065, in clear_search
 File "C:\Users\Konstantin\AppData\Roaming\pyRevit\pyRevit\develop\pyrevitlib\pyrevit\forms\__init__.py", line 1041, in search_txt_changed
 File "C:\Users\Konstantin\AppData\Roaming\pyRevit\pyRevit\develop\pyrevitlib\pyrevit\forms\__init__.py", line 932, in _list_options
 File "C:\Users\Konstantin\AppData\Roaming\pyRevit\pyRevit\develop\pyrevitlib\pyrevit\forms\__init__.py", line 1860, in name
Exception: This operation is not valid for when the revision numbering is per sheet.

Script Executor Traceback:
Autodesk.Revit.Exceptions.InvalidOperationException: This operation is not valid for when the revision numbering is per sheet.
[...]

PR #1781, opened 2023-04-03
#1705

nodatasheet:
This PR is to address the issue  #1705


jmcouffin:
thanks for tis one @nodatasheet 
Like mentioned by @thumDer there might be quite a few other instances of the same issue with the `hasattr`

nodatasheet:
Hey guys!

The problem is that `hasattr` is quite a popular function and rough search gives us ~460 results in the project 😝 

image

Any hint on how to filter the one we're after?

@jmcouffin @thumDer

jmcouffin:
@nodatasheet yep, I would concentrate on the use within the pyrevitlib section and in pyrevit tools and not so much in the site-packages section 
 for now

PR #1785, opened 2023-04-05
Feature change/ReValue {a} -> {{a}}

Addamsovka:
Hi Ehsan,
I have one suggestion for change with the ReValue tool.

It would be nice to use the whole capability of Regular Expression there. With one curly braces, it is impossible to fully use the regex in the fields. The regular expression uses the {number} for the selected amount of elements like this -> ([abcd]){2}.

I found that double curly braces {{ }} could work just fine and allow for full functionality of the regex there.


eirannejad:
Merged into develop and will be released with next version very soon. Thanks @Addamsovka 

There is a format selector now in the UI

Revit_pZrlhBTVLd

Issue #1763, opened 2023-03-16
problem with make pattern tool


kradbim:
Hi,
I've discovered a problem with creating a .pat file by "Make pattern" tool.
I tried to make MODEL pattern, and when tried to import it back to revit to create a custom Model pattern, I received a message that" No MODEL pattern found"
I've investigated the .PAT file and discovered that there is a mistake in the columns as it combines two columns in one.
Here you have an example:
![image](https://user-images.githubusercontent.com/128045505/225607028-09d88a11-f5b0-4a07-98fd-51dfb36f029c.png)

I've tested it several times in 2023ENG of Revit and also 2021 in polish Revit.
Is that something that you're aware of?
I would be grateful for any suggestion of how can I deal with it.
All the best!

elizabethreiter:
Hi i am having this same issue using Pyrevit where after creating a patter it says "No MODEL pattern found". This seems to be a new issue as of late, as I've never had it before. The code writing is BEYOND my capability with the solution above. Can this be fixed in an upgraded pyrevit plugin? HELP!!!

jmcouffin:
Fixed by @alexdaversa in https://github.com/eirannejad/pyRevit/pull/1792
See wip release in the action section

jmcouffin:
@kradbim Look for wip installers in this article
https://pyrevitlabs.notion.site/pyRevit-CI-CD-92078f5fb72440de8e7b63f00fa72725
and you can get it right away

PR #1791, opened 2023-04-07
Fix Issue 1763

alexdaversa:
Fix for issue #1763 where pattern file missing comma when exported


alexdaversa:
Sorry, closing to instead pull onto develop

PR #1792, opened 2023-04-07
#1763 missing comma in exported patterns

alexdaversa:
Fix issue #1763 pattern separator missing from joined pattern string


jmcouffin:
Thanks @alexdaversa

PR #1797, opened 2023-04-12
Added send ExecParams by name

dosymep:
https://github.com/eirannejad/pyRevit/issues/1697


eirannejad:
Merged into develop and will be released with next version very soon. Thanks @dosymep

Introduced a new static assembly called `PyRevitLabs.PyRevit.Runtime.Shared` that can contain shared types and logic from now on. It now contains a readonly `ExecParam` that can be used by invoke bundles:

```csharp
using pyRevitLabs.PyRevit.Runtime.Shared;

    public class PyRevitTestInvokeCommandWithExecParams : IExternalCommand
    {
        public ExecParams execParams;

        public PyRevitTestInvokeCommandWithExecParams() { }

        public Result Execute(ExternalCommandData revit, ref string message, ElementSet elements)
        {
            TaskDialog.Show("Revit", $"Hello World from Invoke Button!!!\nName:\n\"{execParams.CommandName}\"");
            return Result.Succeeded;
        }
    }
```

The dynamically generated runtime links to this library and will use this to set the `execParams` on the invoke bundle class instance. Now both use the same type.

PR #1799, opened 2023-04-14
Russian localization Settings Window

dosymep:
I have added localization for Settings Window.


jmcouffin:
Nice one
@eirannejad Do you see something in particular that would mess something else?
If not, we shall merge it. 
After that I can add the french version eventually

dosymep:
> Nice one @eirannejad Do you see something in particular that would mess something else? If not, we shall merge it. After that I can add the french version eventually

I think need test in others engines, I have tested in IronPython 2.7.

dosymep:
I think I will gradually translate, as this is a difficult task, plus there will be updates for the new revit, and there may be merge conflicts.

I think I will do the translation of windows in different Pull Requests.

dosymep:
in a good way, you need to write a utility that would receive all window translations and compare translation identifiers

eirannejad:
Merged into develop and will be released with next version very soon.  Thanks @dosymep @jmcouffin

PR #1808, opened 2023-04-27
Russian localization buttons

dosymep:
I have added button localizations.


jmcouffin:
Thanks for that. 
Nice of you  @dosymep

PR #1816, opened 2023-05-18
FamilyLoaderOptionsHandler modification

Pete-Smith:
I ran across this when working on my own plug-in. The DB.Document.LoadFamily method wouldn't reload the family types until I made this change.

Some parameters in this interface are passed by-reference (using the out keyword in the argument list).

These references are changed with their Value attribute. The values set in the current implementation don't stick, counter to the intent of the code.

I also set the overwriteParameterValues, for consistency.

See this IronPython explainer. ([Link](https://ironpython.net/documentation/dotnet/dotnet.html#methods-with-ref-or-out-parameters))

Thank you! We love pyRevit at CORE Design Group.


jmcouffin:
@Pete-Smith 
Do you have a specific sample code to give it a quick ride. Or even better, write a unit test in the dev extension!
In the meantime, I will merge into develop
Thanks

Pete-Smith:
I'd be delighted to back this up with some test code.

PR #1825, opened 2023-05-23
Misspelling of Highlight

Pete-Smith:
This is great feature! Very much needed. 
I had an idea for a similar feature, and was using the working title: "Lie Detector".


jmcouffin:
Lie detector... Nice one.
I only use fun names with the tools i build for my team.
Jokes and sarcasm is not really transnational (that's the french in me speaking)

PR #1837, opened 2023-06-10
Fix 2d toggle reload behaviour

jmcouffin:


jmcouffin:
uppon Reload, icon was not found: was loking for the off.png in the reload.pushbutton folder

PR #1852, opened 2023-06-26
Issue #1840

### LINKED ISSUE #1840: Guard stack issues in large list forms
I've been encountering frequent crashing when forming selections in large lists using the form 'forms.SelectFromList'.

I have tried this using elements, custom class to show name/pass element and also just raw strings with same result regardless of the UI key/value types I provide.

The error relates to the guard stack, and crashes out of the model when cancelled or OK'd. I am aware any error relating to this is an indicator of running out of memory, so was wondering if this is either:

A. An issue with the form (e.g. memory leak)
B. A general limitation of the form in pyRevit
C. A general limitation of the form in WPF, generally

If it's just a matter of it being a windows/WPF limit, I'm happy with that as an answer/close out, just want to know what is the likely cause so I can build around or train around it for users as crashing out of sessions in big models is brutal. Generally all other tools/features have been great in large models such as progress bars and cancellation/script.exit() capability that pyRevit offers.

Steps to reproduce the behaviour:
1. Large model (1GB+)
2. Generate a list form with at least 200 elements
3. Shift select enough elements that it visually exceeds the UI (shift select)

In most cases, this will crash the model, but it's not entirely consistent.

I am using Windows 10, Revit 2023 and pyRevit 4.8.12.22247. My machine has 64GB (HP ZBook) and intel corevproi9 processor, so guessing it's potentially virtual memory related. Have tried manually assigning paging size for C and same issue persists (generally we run machines on auto).

I have tested this on current WIP build and have encountered similar issues.

dosymep:
fix #1840

### DIFF
--- a/pyrevitlib/pyrevit/forms/__init__.py
+++ b/pyrevitlib/pyrevit/forms/__init__.py
@@ -758,6 +758,8 @@ class SelectFromList(TemplateUserInputWindow):
 
     """
 
+    in_check = False
+    in_uncheck = False
     xaml_source = 'SelectFromList.xaml'
 
     @property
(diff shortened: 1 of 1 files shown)


jmcouffin:
Will merge this one and try it right away

jmcouffin:
thanks @dosymep

Issue #1830, opened 2023-05-25
REQUEST: Centre of gravity finder


dr3141:
**Is your feature request related to a problem? Please describe.**
Working in precast industry finding a centre of gravity for an assembly is crucial to our day to day work. This can be done with the use of dynamo but it's a bit cumbersome. There used to be a great Revit add-in specifically for this, but  the devs stopped supporting it after Revit 2020.

**Describe the solution you'd like**
A one click solution within Revit for finding centre of gravity, anything similar to what this add-in does on older versions of Revit:
https://apps.autodesk.com/RVT/en/Detail/Index?id=2883284966960053552&appLang=en&os=Win64
The add-in locates the centroid of selected assembly and places three short lines XYZ in the model.

**Describe alternatives you've considered**
Dynamo scripts, alternative add-ins, exporting assemblies to CAD and using AUTOCAD, all work to some extent but not very user friendly, consistent or easy to use.

thumDer:
This could be quite easy to implement, `Solid` class has a `GetCentroid()` or similar method to get the coordinates of the complete element geometries centroid, just keep in mind it is the center of gravity only if the element has one single material only. If that is what you meant it is possible.

dr3141:
> This could be quite easy to implement, `Solid` class has a `GetCentroid()` or similar method to get the coordinates of the complete element geometries centroid, just keep in mind it is the center of gravity only if the element has one single material only. If that is what you meant it is possible.

That's exactly what I meant, coordinates of the centroid for a single material element.

jmcouffin:
resolved by #1857

dr3141:
Thank you both! It works great with families and generic models! 

Is there any chance the three lines that indicate the centre o gravity use a custom line style? For easy distinction from other lines in the model? For example 'CoG line' or something similar?

On a side note, for any potential users, it does not work when selecting an assembly (it produces an error - see below), so you need to make sure to filter assemblies out of your selection. Apart from that, it's great!

IronPython Traceback:
Traceback (most recent call last):
 File "C:\Users\dr\AppData\Roaming\pyRevit-Master\extensions\pyRevitTools.extension\pyRevit.tab\Analysis.panel\Tools.stack\Analyse.pulldown\Get Centroid.pushbutton\script.py", line 59, in <module>
 File "C:\Users\dr\AppData\Roaming\pyRevit-Master\pyrevitlib\pyrevit\revit\db\query.py", line 1579, in get_geometry
[...]

PR #1857, opened 2023-07-03
Issue #1830

thumDer:
New tool based on #1830


jmcouffin:
It looks good 
I will try it out tomorrow or wednesday

jmcouffin:
Thanks @thumDer

extensions/pyRevitTools.extension/pyRevit.tab/Analysis.panel/Tools.stack/Analyse.pulldown/Get Centroid.pushbutton/script.py
```diff
@@ -0,0 +1,94 @@
...
+    Raises:
+        ValueError: If the input is an empty list
+
+    Returns:
+        DB.Solid: The union of the solids
+    """
+    union = None
+    if not solids:
+        raise ValueError('List of provided solids was empty.')
+    for solid in solids:
```
jmcouffin, line 27: You may need a try except somewhere here as it does not process to merging all the solids and when it cannot it throws:
```
File "C:\pyRevit\B1_ToolBar.extension\Outils BIM One.tab\test.panel\test.pulldown\z.pushbutton\script.py", line 49, in <module>
 File "C:\pyRevit\B1_ToolBar.extension\Outils BIM One.tab\test.panel\test.pulldown\z.pushbutton\script.py", line 27, in merge_solids
Exception: Failed to perform the Boolean operation for the two solids. This may be due to geometric inaccuracies in the solids, such as slightly misaligned faces or edges. If so, eliminating the inaccuracies by making sure the solids are accurately aligned may solve the problem. This also may be due to one or both solids having complexities such as more than two faces geometrically meeting along a single edge, or two coincident edges, etc. Eliminating such conditions, or performing a sequence of Boolean operations in an order that avoids such conditions, may solve the problem.
```

jmcouffin: other than that, I love it!

jmcouffin (changes requested):

extensions/pyRevitTools.extension/pyRevit.tab/Analysis.panel/Tools.stack/Analyse.pulldown/Get Centroid.pushbutton/script.py
thumDer: You're right, I've already encountered that issue before in other tools. I think this might be related to the limitation that you can't have edges shorter than ~0.7mm, and the inaccuracies can produce such edges on the joined geometries, so it throws this exception.
I will undo the whole process, and try to provide some details on the problematic elements.

thumDer: I've changed the code, please have a look. Unfortunately the reference extracted from the solid points to the element's type, so reporting the id is not useful...

extensions/pyRevitTools.extension/pyRevit.tab/Analysis.panel/Tools.stack/Analyse.pulldown/Get Centroid.pushbutton/script.py
```diff
@@ -0,0 +1,108 @@
...
+                    'it will be excluded from the calculation'.format(
+                        revit.ElementWrapper(e).name
+                    )
+                )
+    return union
+
+
+selection = revit.get_selection()
+if not selection:
+    forms.alert('You must select one element.', exitscript=True)
```
jmcouffin, line 52: added the error handling when nothing is selected

jmcouffin (approved):

extensions/pyRevitTools.extension/pyRevit.tab/Analysis.panel/Tools.stack/Analyse.pulldown/Get Centroid.pushbutton/script.py
thumDer: or we could just add the selection context to the bundle. would you?

Issue #1749, opened 2023-03-03
CPython interpreter not working


narrsam:
I am trying a execute a very simple script as a test in preparation to using a CPython library in my script:

```
#! python3

import Autodesk.Revit.DB as DB
from Autodesk.Revit.UI import TaskDialog
from pyrevit import revit

doc = __revit__.ActiveUIDocument.Document
td = TaskDialog.Show("testing", doc.PathName)
```

The script works fine when I remove the first line (#! python3), but crashes with this error if I leave it:

```
CPython Traceback:
KeyNotFoundException : KeyError
 at IronPython.Runtime.PythonDictionary.GetItem(Object key)
 at IronPython.Runtime.PythonDictionary.get_Item(Object[] key)

pyRevitLabs.PythonNet
[...]

PR #1862, opened 2023-07-06
fix: make envvars cpython compatible

sanzoghenzo:
first step towards fixing #1749


jmcouffin:
@sanzoghenzo I will merge this one and see if it makes someone else unhappy later :) 
thanks

PR #1863, opened 2023-07-07
Localization Keynotes

dosymep:


jmcouffin:
Great job @dosymep !
Now you got me in a corner and I will have to do the french version!

PR #1865, opened 2023-07-10
handling #1864

### LINKED ISSUE #1864: Failed to save config file
**Describe the bug**
I understand that this is a long time known bug, and it still happens quite offen to our users. If multiple revit sessions are starting simulteanously, some of them fails to load pyrevit, because `pyRevit_config.ini` is locked by one of the others. We should fiugre out something to fix it. Related to #1280 

kép

thumDer:
This is an attempt to fix #1864 .
There might be sideeffects, so @jmcouffin or @eirannejad should have a look on it, if this is sufficient.


jmcouffin:
@thumDer 
I just tried it.
You rock man! 
That little `try except` seems to have done the trick on something that has bothered me for quite a while.

PR #1871, opened 2023-07-17
Option to set icon size for toggle_icon and UIpanel's is_beta

LTusche:
toggle_icon() now allows you to select the icon size, so they don't change when you use them.
Examples:
script.toggle_icon(new_state, icon_size = script.ICON_LARGE)
script.toggle_icon(new_state, icon_size = script.ICON_MEDIUM)
script.toggle_icon(new_state, icon_size = script.ICON_SMALL)

UI Panels can now use "is_beta" in yaml to avoid creating empty panels when all of its buttons are beta buttons.
A cleaner solution without "is_beta" should be made.


LTusche:
@jmcouffin Created this new commit with the corrected source.

jmcouffin:
thanks @LTusche 
you should be able to try the new WIP installers in a bit under the actions tab of github

PR #1872, opened 2023-07-18
config mode for Get Centroid

thumDer:


jmcouffin:
Works on 2024 and 2023 on my side.
But you are right, the implementation of the shift click made by @thumDer is slightly different from the common one, as it does not produce the little `dot `next to the tool button.

jmcouffin:
to trigger the little dot, it takes a config.py file in the same folder to take care of the shift+click process.

dr3141:
You are right, @jmcouffin it works! Apologies for the confusion, @thumDer

A 3D element needs to be selected in the model before shift clicking the Get Centroid icon. I was trying to do it without selecting anything first.

thumDer:
I can move the part handling the config to a `config.py` but that way it is not possible to use the selection context anymore I think. But if that solution make more sense to you I can update it.

PR #1880, opened 2023-07-31
dc3d server implementation with example

thumDer:
This is a pyrevit based implementation of an `IDirectContext3DServer` Interface, that can be used to show low level geometry overlaid on the actual revit geometries.

### DIFF
--- a/pyrevitlib/pyrevit/revit/dc3dserver.py
+++ b/pyrevitlib/pyrevit/revit/dc3dserver.py
@@ -1,0 +1,627 @@
+# -*- coding: utf-8 -*-
+"""DirectContext3DServer wrapper and utilities"""
+
+from collections import namedtuple
+import traceback
+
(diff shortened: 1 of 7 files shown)


pyrevitlib/pyrevit/revit/dc3dserver.py
```diff
@@ -0,0 +1,627 @@
...
+                return
+
+            outline = DB.Outline(vertices[0], vertices[1])
+
+            for idx in range(2, len(vertices)):
+                outline.AddPoint(vertices[idx])
+
+            return outline
+        except:
+            print(traceback.format_exc())
```
thumDer, line 531: I still have to finish this.

thumDer: It is actually finished, nevermind.

PR #1900, opened 2023-08-23
UI theme (toggles)

dosymep:
I have add support themes to toggles.
image


jmcouffin:
Looks good.
Thks @dosymep

PR #1905, opened 2023-08-31
SelectFromList class - allow different group sorting methods

GerhardPaw:
Introduced the optional argument "sort_groups". 
Allowed inputs are "sorted", "unsorted", "natural." Defaults to "sorted". 
A static function for natural sorting was added.
Description and example was added.
"import re" was added.


jmcouffin:
It looks good.
Will merge it. Please do test the function afterwards using the wip installers.
Thks

PR #1910, opened 2023-09-06
pyRevitCore metadata file

dosymep:
I added extension.json file to pyRevitCore


dosymep:
I need this file because my extension controls the visibility of standard pyRevit extensions

PR #1916, opened 2023-09-08
Create worksets_content_check.py

jmcouffin:


jmcouffin:
Check that lists the elements per workset
- removed proxies and irrelevant elements
- added table list of elements (can be lengthy and does not support linkify of ids)

PR #1919, opened 2023-09-14
docs: switch doc generation to mkdocs

sanzoghenzo:
Hi, this is the PR to move to mkdocs!

The automatic publishing is done by `.github/workflows/docs.yml`;
As of now it triggers on pushes to `docs` branch and when a PR is issued on main branch;
should I change it to only trigger on pushes to main branch, or even better only on new releases?

To answer your question from the forum, this is what it needs to be done:

image

- go to settings
- select the page section
- under "Source" select "Deploy from a branch"
- select the "gh-pages" branch (it should be created the first time that the workflow runs) then click save.


jmcouffin:
@sanzoghenzo 🎩 off
Great job.
I will try it on my side

jmcouffin:
Looks good on my fork
I will give it a try on the dev branch then! 🤞

PR #1931, opened 2023-09-29
Fix invokebutton

dosymep:
I fixed a bug with the assembly cache that I introduced in the last PR.

Wrong bundle config

was
image

became
image


dosymep:
This PR added wrong work with InvokeButton.

I don't know what InvokeDLLEngine one for all assemblies, there was also a code that misled me.

https://github.com/eirannejad/pyRevit/pull/1777

dosymep:
Maybe I broke everything 😔

jmcouffin:
Let's try it then

Issue #1913, opened 2023-09-07
Set Crop Region to Shape (Sheets) does not crop Elevations


jmcouffin:
**Is your feature request related to a problem? Please describe.**
Set Crop Region to Shape (Sheets) does not crop Elevations

**Reference**
https://discourse.pyrevitlabs.io/t/set-crop-region-to-selected-shape/2047

PR #1932, opened 2023-09-29
Fix for #1913 - Set Crop Region now works with Elevations

Negazero:
fix for #1913.
The command still has issues with setting crop region for 3D views, and you may have to run the command multiple times to align properly, but these problems appeared to exist with original as well.


jmcouffin:
It looks good @Negazero 
Thanks for taking up the challenge!

Issue #1949, opened 2023-10-19
Error loading pyRevit


chttIE:
Good afternoon.
First of all I would like to thank you for pyrevit. thx!)
![image](https://github.com/eirannejad/pyRevit/assets/128715491/cc7fb052-7970-4ced-a5c1-dc877cf803f7)

The error shown on the screen last time was solved by changing the manifest name to 0Pyrevit.....
But I would like all plugins to just be friends with each other without crutches.

Digging deeper, it became clear that the library traces are a bit behind those used by other plugins, which is why the error appears.
![image](https://github.com/eirannejad/pyRevit/assets/128715491/c338e689-2032-463e-b4b3-960e913957d9)
especially ControlzEx.dll

Tell me if you are going to move to new versions of libraries

jmcouffin:
Thanks for digging in.
Please create a PR to the develop branch with these updated dll version, and we will see how it goes once the CI-CD kicks in

jmcouffin:
@chttIE 
- ControlZEX you are right
- Mahapps is recompiled under pyrevitlabs, so it should not be an issue
- MaterialDesign* is not part of the repo / project, or at least I cannot find it.

jmcouffin:
https://pyrevitlabs.notion.site/Custom-Changes-d9ea2ad6b6174bb4b583e20ab08fb0ef

https://github.com/eirannejad/pyRevit/issues?q=is%3Aissue+MahApps+

Wilionaire:
I have this same issue, followed most of the fixes (pyrevit env + pyrevit doctor in cmd) removed all the add-ins I could. Reinstalled etc etc. Still having the same issue. Appreciate all the hard work from everyone, Im not skilled enough to help. Just to put my hand up with the same issue,

jmcouffin:
@Wilionaire try and prefix the pyRevit.addin file like so 0_pyRevit.addin and report if it works

Wilionaire:
@jmcouffin thanks. In my previous attempts to fix the issue I had uninstalled and reinstalled pyrevit. The error was being presented, but then pyrevit would not load. 

After adding the prefix as suggested, I am still getting the error, but pyrevit is now available which is where we were at the start of this. I tested the prefix across multiple years 20, 21, 22 and 23. The error and pyrevit tab present in all except 22. In 22, the error presented by no pyrevit tab. I hope this helps?

jmcouffin:
updated ControlZeX.dll with #2011

PR #1979, opened 2023-10-24
Change ControlzEx.dll (version 4.4.0.50)

leomalyshev:
Hello. I change ControlzEx.dll, check this #1949 

Could you make strong name for pyRevitLabs.MahAppsMetro.dll? 
Because MahApps.Metro has strong name, but I see mistake


jmcouffin:
Is there a reason why you are bringing the 4.3.0 version when the 6.0.0 version is available? Incompatibility?

jmcouffin:
And looking at files and files content mentioning Controlzex, there is more to it just replacing the dll, 
Can try and compile the solution on your side before I push the PR in the ci/cd? @leomalyshev

jmcouffin:
@leomalyshev 

> Is there a reason why you are bringing the 4.4.0.50 version when the 6.0.0 version is available? Incompatibility?

jmcouffin:
updated to ControlZeX v6.0 for .NET framework v4.0 with #2011

jmcouffin:
Add to rollback to the old version as it was creating issues and requires extensive work nobody is in the capacity of doing right now.
see #2012

Issue #1984, opened 2023-11-01
Visual Studio 2022 build fails when loading plugins / commands w/o AddInId


MYu1Gpzumr:
**Describe the bug**
Visual Studio fails to launch Revit 2023 when pyRevit is installed, hitting an error in the following class:
```
using System.Xml;
using pyRevitLabs.Common.Extensions;

namespace pyRevitLabs.TargetApps.Revit
{
	public class RevitAddonManifest
	{
		public string FilePath { get; set; }

		public string Name { get; set; }

		public string Assembly { get; set; }

		public string AddInId { get; set; }

		public string FullClassName { get; set; }

		public string VendorId { get; set; }

		public RevitAddonManifest(string manifestFile)
		{
			FilePath = manifestFile;
			XmlDocument xmlDocument = new XmlDocument();
			xmlDocument.Load(manifestFile);
			Name = xmlDocument.DocumentElement.SelectSingleNode("/RevitAddIns/AddIn/Name").InnerText;
			Assembly = xmlDocument.DocumentElement.SelectSingleNode("/RevitAddIns/AddIn/Assembly").InnerText.NormalizeAsPath();
			AddInId = xmlDocument.DocumentElement.SelectSingleNode("/RevitAddIns/AddIn/AddInId").InnerText;
			FullClassName = xmlDocument.DocumentElement.SelectSingleNode("/RevitAddIns/AddIn/FullClassName").InnerText;
			VendorId = xmlDocument.DocumentElement.SelectSingleNode("/RevitAddIns/AddIn/VendorId").InnerText;
		}
	}
}

```

where `AddInId = xmlDocument.DocumentElement.SelectSingleNode("/RevitAddins/Addin/AddinId").InnerText;` reports a **System.NullReferenceException**, determined via inspection, for plugins that have a `ClientId`, but not an `AddInId`. One such addin is the Autodesk.BatchPrint.addin, found @ C:\ProgramData\Autodesk\Revit\Addins\2020 for me.

Autodesk.BatchPrint.addin
```
<?xml version="1.0" encoding="utf-8"?>
<RevitAddIns>
  <AddIn Type="Application">
    <Name>Batch Print Application</Name>
    <Assembly>C:\Program Files\Autodesk\Batch Print for Revit 2020\BatchPrint.dll</Assembly>
    <ClientId>9d508589-48b6-4ff6-a50d-d16d77c585a1</ClientId>
[...]

PR #1990, opened 2023-11-07
Accept ClientId when loading plugins

MYu1Gpzumr:
A fix for #1984 

Accepts ClientId in place of AddInId, which allows older plugins to be loaded w/o error when Revit is launched from Visual Studio.


jmcouffin:
@MYu1Gpzumr thanks. But before merging, I think it requires building the dll?

PR #2010, opened 2023-11-22
fix for CI/CD workflows

jmcouffin:
updated actions


jmcouffin:
change minimal version of visual studio sln
# Visual Studio Version 17
VisualStudioVersion = 17.7.34003.232
MinimumVisualStudioVersion = 10.0.40219.1

PR #2014, opened 2023-12-02
improved view handling

thumDer:
Added an option to reopen the closed views after synchronization

### DIFF
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/sync.pushbutton/config.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/sync.pushbutton/config.py
@@ -6,13 +6,29 @@ doc = revit.doc
 
 def close_inactive_views():
 
-    form = forms.SelectFromList.show(["Yes", "No"], title = "Close all inactive views before synchronisation", height = 200)
-    if form=="Yes": 
-        setattr(my_config, "close_inactive_views", True)
-        script.save_config()
-    else:
-        setattr(my_config, "close_inactive_views", False)
-        script.save_config()
+    try:
+        current_setting = getattr(my_config, "view_handling")
+    except:
+        current_setting = "nothing"
+    options = {
+        "nothing": "Don't touch my views",
+        "reopen": "Close them, but reopen them right after",
+        "close": "Close them, and call it a day"
+    }
+
+    selection = forms.ask_for_one_item(
+        items=options.values(),
+        default=options.get(current_setting),
+        title="View Handling",
+        prompt="Specify what to do with open views"
+    )
+
+    setattr(
+        my_config,
+        "view_handling",
+        next((k for k in options if options[k] == selection), None)
+    )
+
 
 if __name__ == "__main__":
     close_inactive_views()
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/sync.pushbutton/script.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/sync.pushbutton/script.py
@@ -4,46 +4,93 @@ from pyrevit import script, forms, coreutils
 
 doc = revit.doc
 
+logger = script.get_logger()
+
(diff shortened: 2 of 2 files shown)


extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/sync.pushbutton/config.py
```diff
@@ -6,13 +6,29 @@
...
+        title="View Handling",
+        prompt="Specify what to do with open views"
+    )
+
+    setattr(
+        my_config,
+        "view_handling",
+        next((k for k in options if options[k] == selection), None)
+    )
+
```
jmcouffin, line 32: use `script.save_config()` to save to pyRevit-config.ini

extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/sync.pushbutton/script.py
```diff
@@ -4,46 +4,93 @@
...
+logger = script.get_logger()
+
 my_config = script.get_config()
 try:
-    close = getattr(my_config, "close_inactive_views")
+    view_handling = getattr(my_config, "view_handling")
 except:
-    close = False
+    view_handling = "nothing"
+
```
jmcouffin, line 20: I would `setattr `than use `script.save_config()` to have a base config after first run of the tool

extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/sync.pushbutton/script.py
```diff
@@ -4,46 +4,93 @@
...
+    if starting_view:
+        uidoc = revit.uidoc
+        HOST_APP.uidoc.RequestViewChange(starting_view)
+        uidoc.ActiveView = starting_view
+        all_open_views = uidoc.GetOpenUIViews()
+        for ui_view in all_open_views:
+            if view_handling == "reopen":
+                view_cache.append(ui_view.ViewId)
+            doc_view = document.GetElement(ui_view.ViewId)
+            if doc_view.Name != starting_view.Name :
```
jmcouffin: While trying it on a dummy file with view names identical: Level 1, it did not close anything.
Comparing Id will be better as you can have views with identical names.
`if doc_view.Id != starting_view.Id :`

extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/sync.pushbutton/config.py
thumDer: That's strange, for me it remembered the previous setting without saving... Maybe that one only really matters between different Revit sessions? Anyway, I'm adding it.

extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/sync.pushbutton/script.py
thumDer: I've checked this, and if you have 2 different documents with the same named view open this is intended, the `uidoc.GetOpenUIViews()` returns only views from the active document so the view with the same name but from a different document is a different context. View names are unique per document, so the name comparison should be sufficient. I think the reason for nothing has happened is that you have no starting view specified in your dummy project, so an info balloon should have showed up saying "No Starting View Set". Could you double check this?

jmcouffin: Checked.
When you have for example: Level 1 (structural plan) and Level 1 ('regular' plan and being the starting view) it does not close either as the names are indeed the same (Id and types are different)

thumDer: Damn, you are right, I almost never have structural plans :D I'll handle it.

jmcouffin (approved):

jmcouffin:
thanks @thumDer

PR #2021, opened 2023-12-18
Color splasher from BIMOne inc. - port from Nonica to pyRevit

jmcouffin:
Color Splasher allows you to color the elements of the active view according to their parameter values. It also allows the creation of a legend or a list of view filters.

- Based on ColorSplasher by BIMOne - MIT License - 2021 BIM One Inc.
- First Dynamo - Python edition December 2023: Nonica by Estudio Alonso Candau SLP - Jaime Alonso Candau - See specific non-commercial license agreement for Nonica in the script
- Refactored, pyRevit-ed and added View Filters Creation version by Jean-Marc Couffin, December 2023


jmcouffin:
work from extension but not from pyrevit main set of tools

PR #2082, opened 2024-02-10
refactor: remove compiled assets from bin folder

sanzoghenzo:
Following the discussion with @jmcouffin, this PR removes all the files that are being compiled by the `pipenv run pyrevit build products` step in the CI.

I left the bin/engines folder untouched.
Also, there are some images/icons and json files in the bin folder that are not copied from other sources, so I've kept them there.
I edited the .gitignore file to avoid further additions in the bin folder (with the exception of the aforementioned files and subdirectory).

Please note that I didn't test this with revit, I had to do this at home and I don't have a personal license to use to test.
In other words, please test this before merging!


jmcouffin:
Nice @sanzoghenzo 
I will try it once the CI/CD is done with it. 
I invite you to try it as well.

jmcouffin:
Tried and approved. 🚀

PR #2085, opened 2024-02-11
Stale issues bot

sanzoghenzo:
This will add a bot to automatically mark as stale issues older than 60 days, and close them after 14 days if there is no activity.

This is an attempt to prevent the buildup of issues that no longer apply, and avoid having to check and close them manually as @jmcouffin had to do a while ago.

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -1,0 +1,20 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v9.0.0
+        with:
+          days-before-issue-stale: 60
+          days-before-issue-close: 14
+          stale-issue-message: "This issue is stale because it has been open for 60 days with no activity."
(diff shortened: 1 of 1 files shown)


jmcouffin:
@sanzoghenzo 
The checks failed. Do you think this is a false positive?
Waiting for your go to merge.

sanzoghenzo:
This PR doesn't alter the source, so we should be good to go.

To prevent these errors, we should change the workflows to only trigger on source code changes.

I honestly don't know what happened in the build job...

jmcouffin:
Do you happened to know if it will work on actual issues or just the new ones and on?

sanzoghenzo:
> Do you happened to know if it will work on actual issues or just the new ones and on?

It runs on all the issues; there's the [start-date](https://github.com/actions/stale?tab=readme-ov-file#start-date) option to choose the date before which ignore the issues, if you want.

jmcouffin:
I will apply your changes directly to the main branch as well. So that it works right away.

jmcouffin:
added to the master branch (I did not look into it enough to figure out if it runs whatever the branch it is on, but guessing this has to be on the master one)

sanzoghenzo:
Awesome,I can see the (empty) workflow in the actions section. It is scheduled to run every night at 1:30 (i suppose it is UTC)

PR #2086, opened 2024-02-12
Run CI workflow only official repository

sanzoghenzo:
This adds a condition on the github workflow to only run on the main repository, so that the forks don't have to run and fail the workflow due to the signing related job steps.


jmcouffin:
It makes sense, thks

jmcouffin:
It seems to work fine on my own fork. Thanks Again

PR #2094, opened 2024-02-17
Drop old ipy engines

sanzoghenzo:
This removes the old IronPython engines, leaving only 2.7.11 and 3.4.0.

Note that building the installer fails due to the missing runtime csharp files (moved by 584eda7c83563c81f631bc4dfb79a2267f2317bf).

I also added a workflow to do the build only (`pyrevit build products`) on forks (so that we can ensure everything compiles before submitting a PR).


jmcouffin:
You want to use which base branch @sanzoghenzo ?
You may want to rebase?

sanzoghenzo:
Thanks for the pointer @jmcouffin, I always forget to set the right branch :sweat_smile: 

I did the work on the develop branch, should I do it on the 4x one?

jmcouffin:
I'd say both, yes please

jmcouffin:
Would you mind trying the installer for the dev 4 branch? @sanzoghenzo

PR #2095, opened 2024-02-18
drop old ironpython engines

sanzoghenzo:
Only 2.7.11 and 3.4.0 versions will be available. The default version is 2711


sanzoghenzo:
As requested in the other PR, I built the installer locally and everything seems to work fine 😉

PR #2105, opened 2024-02-21
fix: recursion on emojis renderer

### LINKED ISSUE #2104: Extra characters on repeated Emojis
# 🐞 Describe the bug

Problem raised in [this forum discussion](https://discourse.pyrevitlabs.io/t/emojis-breaking-output-print-table/2667).

image

Note that this is not strictly related to tables, but with repeated emojis and how the [`Emojize` c# method performs the substitutions](https://github.com/eirannejad/pyRevit/blob/6f72bb9b3b5389d281a2ad1eba2995e94be32d9b/dev/pyRevitLabs/pyRevitLabs.Emojis/Emojis.cs#L2650-L2670).

The regex `Match` finds all the shorthands, then for each of them it retrieves the icon and performs a  replace on the entire text with an html  tag with the shorthand as title.
The subsequent match of the same shorthand repeats the replace on the shorthand contained in the title, causing a recursion. that produces extra `">`s in the output.

The solution here is to take only the unique/distinct matches of the regex.

As I discussed in the linked topic, it may also be the case to drop the zip archive containing the icons and rely on fonts such as [Noto Color Emoji](https://fonts.google.com/noto/specimen/Noto+Color+Emoji) and unicode chars.

I'm already working to address both issues.

sanzoghenzo:
fixes #2104 

Since this doesn't depend on Revit APIs, I added a unit test project that could be inserted in the CI pipeline (I will add it soon).

This still uses png icons, I'll tackle the unicode chars / emojis font in a separate PR.

This could be cherry picked in devel branch without any problem

### DIFF
--- a/dev/pyRevitLabs/pyRevitLabs.Emojis.Tests/pyRevitLabs.Emojis.Tests.cs
+++ b/dev/pyRevitLabs/pyRevitLabs.Emojis.Tests/pyRevitLabs.Emojis.Tests.cs
@@ -1,0 +1,25 @@
+using System;
+using System.Text.RegularExpressions;
+using NUnit.Framework;
+using pyRevitLabs.Emojis;
+
+namespace pyRevitLabs.Emojis.Tests
+{
+    public class Tests
+    {
+        [SetUp]
+        public void Setup()
+        {
+        }
+
+        [Test]
+        public void TestRepeatedEmojis()
+        {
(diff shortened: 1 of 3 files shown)


jmcouffin:
It looks good.
Do you want to add the CI step in this PR?

sanzoghenzo:
Since there's only one test, it's too soon for a test step in the CI pipeline to be useful, so I'll defer it for another time

Issue #2109, opened 2024-02-21
Match Tool Stopped Working


Pyrevit-User:
The "Match" tool in Pyrevit has stopped working. I tried uninstalling Pyrevit and re-installing the latest version V.4.8.14, thinking that it could have been a glitch in the script that might get resolved by reinstalling latest. The issue still persists in the new version. This was the tool that enables a Revit user to match the graphic overrides of one source element with other elements easily and quickly, which was a big time saver. Can other users please check and advise on how to solve this issue?
Thank you!

sanzoghenzo:
Hi @Pyrevit-User, you completely ignored te bug report template, in which there were instruction on how to correctly report any iussues. It is difficult to diagnose your problem without more detailed info, for example: 

- what do you mean by "Stopped working"? mothing happens when you click on the button? or it starts and then outputs an error? which error? 
- How can I reproduce the problem? what are the steps involved?
- What are the specs of your environment? (that's the `pyrevit env` part at the end of the template)

Pyrevit-User:
Apologies about the template and sorry for missing it.
I'm only a user not a developer, so I'm not sure what terms like "mothing" and  "specs of environment" are. There is no error that pops when I use the match tool. After I click the match button, the screen gets in the orange mode and directs me to select the source element which I do, then to select the element(s) I want to match, which I also do. Then nothing happens! No matching and no error either.
Thank you!

sanzoghenzo:
"mothing" was a typo, I meant "nothing" 😅 

So the tool runs but it doesn't actually do what is intended to do. This is much more useful than "stopped working" 😉 

Hold the CTRL keyboard key while clicking on the match tool; an output window with many debug messages will popup (and will slow down the loading of the tool, but this is normal).
Leave the window open and use the tool as you described; there may be new messages in the window.
Click on the floppy disk icon in the title bar to save the contents of the window to a file and upload it here.

Thanks in advance for you collaboration!

Pyrevit-User:
Thank you for your guidance. I uploaded the file here to this message.
[Test.zip](https://github.com/eirannejad/pyRevit/files/14376739/Test.zip)

sanzoghenzo:
Thanks, I thought this would be more verbose, here's the content:

```
DEBUG [pyrevit.revit.selection] Picking elements: Element message: multiple: False world: False
[...]

PR #2123, opened 2024-03-02
fix: don't swallow _pick_obj exceptions

sanzoghenzo:
This stems from the investigations done for #2109 

The `selection._pick_obj` function swallows any kind of exception and doesn't inform the user about it.

This fixed it by only catching a specific, expected exception (when the user cancels the selection) and adds a debug log message for it.


jmcouffin:
it looks good
thanks @sanzoghenzo

PR #2131, opened 2024-03-03
use form bug report

sanzoghenzo:
With this we ensure that the requested info are there.

You can see a preview [here](https://github.com/sanzoghenzo/pyRevit/issues/new?assignees=&labels=Bug&projects=&template=BUG_REPORT.yml&title=%5BBug%5D%3A+)


jmcouffin:
all good.
I made the changes directly in the master branch in order to avoid the CI trigger and creating a new major release.
thanks @sanzoghenzo

PR #2135, opened 2024-03-04
Joris van der Meulen > patch forms init

3Pinter:
Some elements (like parts) don't have types. Forms.select_parameters will crash upon feeding such elements.

([forum post](https://discourse.pyrevitlabs.io/t/forms-select-parameters-with-parts/2688))


jmcouffin:
Good to go!
Well done @3Pinter

PR #2137, opened 2024-03-04
fix: reorder search paths

sanzoghenzo:
this will ensure cpython built-in libraries are used instead of pyrevit site-packages

Fixes the issue raised in [this discussion](https://discourse.pyrevitlabs.io/t/how-to-use-python-package-like-langchain/2673) and [this one](https://discourse.pyrevitlabs.io/t/exporting-graphes-chart-wrapper-object-and-3rd-party/1617).

To check that it works:

- create a python 3.8.5 environment and install a package in it, for example langchain
- set the environment path as PYTHONPATH or use `sys.path.append`
- `import langchain`


jmcouffin:
Nice one @sanzoghenzo

PR #2140, opened 2024-03-06
stale issues workflow and bug report template modifications

jmcouffin:
https://discourse.pyrevitlabs.io/t/change-stale-issues-bot-message/2747
@sanzoghenzo


sanzoghenzo:
Forgive me if I sound pedantic, but maybe it's better to move the indication of the debug run in the description of the "Describe the bug" section.

Something like "If the bug is about a pyRevit tool or python scripts, run it in debug mode (CTRL+click on the button) and copy the contents of the output window here."

image

Or should we add a new section ("Error message") with the same kind of the pyrevit env section (`textarea` with `render: shell`)?

jmcouffin:
No worries @sanzoghenzo 
You ask and I shall provide ;p
It is done, I added a section.
Both in the master and dev4 branches

PR #2145, opened 2024-03-09
Ci optimizations

sanzoghenzo:
- The main CI only runs if there are changes inside the source code folders
- The docs CI runs if the documentation or the pyrevitlib folder is changed
- Added a workflow to manage the `needs-more-info` label (automatically added if one of us comments on an issue, automaticalli removed if the author responds) - this works in tandem with the stale bot (that only works on issues labeled with `needs-more-info`.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,10 +4,20 @@ on:
   push:
     branches:
       - docs
+    paths:
+      - 'mkdocs.yml'
+      - 'docs/**'
+      - 'pyrevitlib/**'
+      - '*.md'
   # when PR from develop->master is created
   pull_request:
     branches:
       - master
+    paths:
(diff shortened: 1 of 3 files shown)


jmcouffin:
Pretty cool.
I did the changes manually directly in the master branch for the main.yml file to make sure it does not trigger a release and then rebase your PR on develop-4 . And then merged into master. A bit of an ouroboro situation.
But hopefully with your changes this will not happen again.

sanzoghenzo:
> Pretty cool. I did the changes manually directly in the master branch for the main.yml file to make sure it does not trigger a release . And then merged into master. A bit of an ouroboro situation. But hopefully with your changes this will not happen again.

I think it would have been ok to just merge this pr, but I understand your concern, better be on the safe side 😉

PR #2151, opened 2024-03-10
WIP add cpython 3.12.2

sanzoghenzo:
First attempt at updating python.net to 3.0.3 and adding CPython 3.12.2.

It needs pyrevitlabs/pythonnet PR to be merged, and the submodule to be updated.

I've modified the source to only use the public methods and classes, to keep the edits of the submodule at a minimum and eventually get rid of the fork and use the nuget package directly.

Python.net now needs the path to the python3xx.dll file, I've started ti add the path discovery to the `CPythonEngine` class, but it's  broken (I still have to figure out how can I debug the dynamically compiled code) and hardcoded for the last part (since pyrevit uses an integer to handle the python versions, it's impossible to extract the major and minor version with 100% certainty).


jmcouffin:
Let me know when you feel it is ready to go.
@sanzoghenzo
Same goes for the pythonnet PR

sanzoghenzo:
@eirannejad thoughts convinced me to hold on this PR and wait for the version 5.
Closing this.

PR #2214, opened 2024-04-22
Fix pyrevit releases CLI command and handle naming variations

acco-jpitts:
The ```pyrevit releases``` command has been broken for a while but I was finally able to get it working. 

The initial issue was the check for GITHUBTOKEN env var,

per Github docs: 
[This endpoint can be used without authentication or the aforementioned permissions if only public resources are requested.](https://docs.github.com/en/rest/releases/releases)

So I just commented out that whole block. Then I discovered that due to the addition of multiple different exe and nupkg files, it was trying to download only the first file from each release when you tried to download the installer, so you would get the nupkg. 

The string formatting for the download URL could probably be handled more gracefully, but it works.

I tested ```pyrevit releases download installer <search pattern>``` from v4.8.8 (first installer) - Current and it successfully downloaded the "*_signed.exe" installer for all of them. 

@jmcouffin is "*_signed.exe" non CLI installer the correct one to download this way?

```--pre``` and ```--notes``` flags are both not working still.
and ```latest``` will not find a match either. you have to manually enter 4.8.x for it to locate the tag

I had to change these two lines from False to True in Common.csproj and pyRevitCLI.csproj in order to build successfully, but this is not included in this commit.
```
<Reference Include="pyRevitLabs.Json" HintPath="$(PyRevitDevLibsDir)\pyRevitLabs.Json.dll" Private="True" />
<Reference Include="pyRevitLabs.NLog" HintPath="$(PyRevitDevLibsDir)\pyRevitLabs.NLog.dll" Private="True" />
```


jmcouffin:
Interesting. 
And good findings.

I will merge this one to try it out.
I will also create an issue with the remaining items that require fixing.

PR #2269, opened 2024-05-23
Fix #2268

### LINKED ISSUE #2268: [Bug]: pyrevit Transaction crashes Revit 2024 when canceling occurred failure message
### ✈ Pre-Flight checks

### 🐞 Describe the bug

When using `pyrevit.revit.Transaction()` with default failure handling options `swallow_errors=False`, occurred failure crashes Revit 2024 if you press "Cancel" button.

### ⌨ Error/Debug Message

```shell

```

### ♻️ To Reproduce

Open Revit 2024 (any release including latest 2024.2.1 atm ) and create new project.
To provoke the failure message with the code:
1. Create the wall and copy its id
2. Add project length parameter "P" for walls
3. Add the wall to group
4. Copy the group, so you have two group instances
5. Set parameter using code:
```python
from pyrevit import DB, revit

doc = revit.doc
elem = doc.GetElement(DB.ElementId(1594015))  # get wall by id

with revit.Transaction('Set P', doc):
    elem.LookupParameter('P').Set(1)
```
6. Transaction is completed but then appears failure message. If you cancel it, Revit crashes.
  image

### ⏲️ Expected behavior

Pressing cancel on failure message should cancel the transaction.

```shell
Revit 24.2.1 | Version: 24.2.10.64 | Build: 20240408_1515(x64) | Language: 1033 | Path: "C:\Program Files\Autodesk\Revit 2024\"

mypyrevit | Branch: "master" | Version: "4.8.16:13e75bf" | 
pyRevit CLI v4.8.16.24121+2117.23a0b8cbc4babdeb29a0611e159d017bbe9cbad5
```

### Additional context

I don't have 2023 Revit installed, but in 2022 there is no such behavior. Issue occurs in Revit 2024 only.

nodatasheet:
This PR is to address issue #2268

### DIFF
--- a/pyrevitlib/pyrevit/revit/db/transaction.py
+++ b/pyrevitlib/pyrevit/revit/db/transaction.py
@@ -58,7 +58,7 @@ class Transaction():
                     self._fhndlr_ops.SetFailuresPreprocessor(
                         failure.FailureSwallower()
                         )
-                self._rvtxn.SetFailureHandlingOptions(self._fhndlr_ops)
+            self._rvtxn.SetFailureHandlingOptions(self._fhndlr_ops)
         self._logerror = log_errors
 
     def __enter__(self):
@@ -79,6 +79,7 @@ class Transaction():
                 self._rvtxn.RollBack()
                 mlogger.error('Error in Transaction Commit. '
                               'Rolling back changes. | %s', errmsg)
+        self._rvtxn.Dispose()
 
     @property
     def name(self):
@@ -105,6 +106,7 @@ class DryTransaction(Transaction):
     """Wrapper to a transaction that doesn't commit anything (dry-run)."""
     def __exit__(self, exception, exception_value, traceback):
         self._rvtxn.RollBack()
+        self._rvtxn.Dispose()
 
 
 class TransactionGroup():
@@ -137,6 +139,7 @@ class TransactionGroup():
                 self._rvtxn_grp.RollBack()
                 mlogger.error('Error in TransactionGroup Commit: rolled back.')
                 mlogger.error('Error: %s', errmsg)
+        self._rvtxn_grp.Dispose()
 
     @property
     def name(self):
--- a/pyrevitlib/pyrevit/revit/db/transaction.py
+++ b/pyrevitlib/pyrevit/revit/db/transaction.py
@@ -58,7 +58,7 @@ class Transaction():
                     self._fhndlr_ops.SetFailuresPreprocessor(
                         failure.FailureSwallower()
                         )
-            self._rvtxn.SetFailureHandlingOptions(self._fhndlr_ops)
+                self._rvtxn.SetFailureHandlingOptions(self._fhndlr_ops)
         self._logerror = log_errors
 
     def __enter__(self):


sanzoghenzo:
If what you say in https://github.com/pyrevitlabs/pyRevit/issues/2268#issuecomment-2127536184 is true, then there's no need to get the FailureHandlingOptions and set the options in case of a false-y `swallow_errors`, so you might as well move everything inside that `if swallow_errors`.

And while you're at it, you could drop the `self._fhndlr_ops`, since it is only used inside that method, and use local variable instead.
Also, method chaining could make this a little bit more readable

```python
            if swallow_errors:
                fhndlr_ops = (
                    self._rvtxn.GetFailureHandlingOptions()
                    .SetClearAfterRollback(clear_after_rollback)
                    .SetForcedModalHandling(show_error_dialog)
                    .SetFailuresPreprocessor(failure.FailureSwallower())
                )
                self._rvtxn.SetFailureHandlingOptions(fhndlr_ops)
```

This is just by looking at the code and not running/testing it, so I might have miss something!

nodatasheet:
I agree that moving entire `FailureHandlingOptions` under `swallow_errors` block would totally make sense, but I wanted to make as less changes as possible.
 
`self._rvtxn` is an instance variable. I do not know the reason behind that, but if it is used in some sub-types, there is a risk of breaking someone's logic. They expect it to be `FailureHandlingOptions` no matter what but will be `None` if `swallow_errors is False`.

I followed principle "If it works, don't touch it" 😄

sanzoghenzo:
I see your points, they all make sense (even though I suspect that the use of that private attribute is a simple code smell, and subclasses shouldn't directly access it, but I agree on the principle you followed 😉 )

I still wonder why the code was like that.
Did you test all the permutations of `clear_after_rollback`,`show_error_dialog` and `swallow_errors`?

nodatasheet:
Before PR I checked all of them with the normal failure and they did perform as expected for the proposed change. 

But now I remembered there is a non-sever failure (just a warning that can be ignored). 
For the example I did in #2268 we can keep only one instance of the group. Then Revit will allow the change.

If we set `show_error_dialog` to `False`, it will be shown as a yellow warning at the bottom right of the screen:
image

If we set it to `True` or not set it at all (as it will be for the proposed change), that failure will show a modal window. 
image

So looks like we cannot just hide the setter under the `if` statement as that warning kind of message is what other users expect to see in this case. I will try to see other solutions and come back to this thread. Any ideas are highly appreciated.

nodatasheet:
Follow up on this topic. My further observations show that `SetForcedModalHandling(False)` is the one that causes the issue. So far I found only one workaround.

We can create a preprocessor:
```python
class _SevereFailureDialogChanger(DB.IFailuresPreprocessor):
    """
    Preprocessor that changes failure dialog to modal
    if any of the failures is severe.
    """

    def PreprocessFailures(self, accessor):
        # type: (DB.FailuresAccessor) -> DB.FailureProcessingResult
        NON_SEVERE_FAILURES = (
            DB.FailureSeverity.Warning,
            DB.FailureSeverity['None']
        )
        for msg in accessor.GetFailureMessages():
            if msg.GetSeverity() not in NON_SEVERE_FAILURES:
                opts = accessor.GetFailureHandlingOptions()
                opts.SetForcedModalHandling(True)
                accessor.SetFailureHandlingOptions(opts)
                return DB.FailureProcessingResult.Continue
        return DB.FailureProcessingResult.Continue
```

Then set it to `FailureHandlingOptions` in transaction constructor:
```python
    def __init__(self, name=None,
                 doc=None,
                 clear_after_rollback=False,
                 show_error_dialog=False,
                 swallow_errors=False,
                 log_errors=True,
                 nested=False):
        doc = doc or DOCS.doc
        # create nested transaction if one is already open
        if doc.IsModifiable or nested:
            self._rvtxn = \
                DB.SubTransaction(doc)
        else:
            self._rvtxn = \
                DB.Transaction(doc, name if name else DEFAULT_TRANSACTION_NAME)
            self._fhndlr_ops = self._rvtxn.GetFailureHandlingOptions()
            self._fhndlr_ops = \
                self._fhndlr_ops.SetClearAfterRollback(clear_after_rollback)
            self._fhndlr_ops = \
                self._fhndlr_ops.SetForcedModalHandling(show_error_dialog)

            # THIS IS NEW!
            if not show_error_dialog:
                # catching severe failures and setting error dialog to modal
                dialog_changer = _SevereFailureDialogChanger()
                self._fhndlr_ops.SetFailuresPreprocessor(dialog_changer)

            if swallow_errors:
                self._fhndlr_ops = \
                    self._fhndlr_ops.SetFailuresPreprocessor(
                        failure.FailureSwallower()
                        )
            self._rvtxn.SetFailureHandlingOptions(self._fhndlr_ops)
        self._logerror = log_errors
```
I don't like it as it may slow down the transaction. Also there are potential conflicts with `Application.FailuresProcessing` event handlers (not sure about that). But I don't know other ways.

What I checked:
1. Issue occurs with current PyRevit (IPY340PR and IPY2711) in RVT2024.
1. Issue occurs with PyRevit 4.8.14.24016 (IPY277) in RVT2024.
1. There is no such issue with current PyRevit (IPY2711) in RVT2022.
1. There is no such issue with RevitPythonSell (IronPython 3.4.1) in RVT2024.
1. There is no such issue with Dynamo (IronPython 2.7) in RVT2024, but `SetForcedModalHandling(False)` is ignored for warnings. They are modal anyway.
1. There is no such issue if we don't overwrite `__enter__` and `__exit__`. This works fine with current PyRevit in RVT2024:
    ```python
    with DB.Transaction(doc, 'Set P') as tr:
        opts = tr.GetFailureHandlingOptions()
        opts.SetForcedModalHandling(False)
        tr.SetFailureHandlingOptions(opts)
        tr.Start()
        elem.LookupParameter('P').Set(1)
        tr.Commit()
    ```

I checked other minor things that don't worth mentioning.

I would also appreciate if anyone reproduces the issue in RVT2024 and confirms it exists. I checked on 3 PCs but who knows, maybe I have some common configs.

Since no one else reported it I would keep it as it is.

sanzoghenzo:
Looks good to me, but I don't have the chance to test it.
@jmcouffin care to try this?

jmcouffin:
Good to go, good job @nodatasheet and thanks @sanzoghenzo, good job reviewing and pointing, as always!

PR #2308, opened 2024-07-08
Update convert_icon_to_dark_mode.py

mawdesign:
Fixed issue with pallette based icons by converting to RGBA, and added check for existing dark icon so as to not overwrite ones already specifically made.


jmcouffin:
thanks @mawdesign
Keep them coming!

Issue #2300, opened 2024-07-01
[Bug]: transaction displays errors even if log_errors set to false


kirsch33:
### ✈ Pre-Flight checks

- [X] I **don't have SentinelOne** antivirus installed (see above for the solution)
- [X] I have **searched in the issues** (open and closed) but couldn't find a similar issue
- [X] I have **searched in the [pyRevit Forum](https://discourse.pyrevitlabs.io)** for similar issues
- [X] I already **followed the [installation troubleshooting guide](https://pyrevitlabs.notion.site/Installation-issues-f6495adb56254e9e8021f8a68e3b2ab7) thoroughly**
- [X] I am using the **latest pyRevit Version**

### 🐞 Describe the bug

i am using the pyRevit transaction wrapper and for certain tools I do not want end users to ever see error messages even if the transaction fails. i have set `log_errors=False` but sometimes the logger window still displays the error message. i tracked down the issue to maybe be in this code block:

https://github.com/pyrevitlabs/pyRevit/blob/f9fa0d296386d548e198324dd752fb7fb5cd7b58/pyrevitlib/pyrevit/revit/db/transaction.py#L80

https://github.com/pyrevitlabs/pyRevit/blob/f9fa0d296386d548e198324dd752fb7fb5cd7b58/pyrevitlib/pyrevit/revit/db/transaction.py#L138

it should check if `log_errors` is true before logging this error like it does a few lines up? Or is this by design and I'm mis-understanding the intention of `log_errors`?

### ⌨ Error/Debug Message

```shell
not applicable
```

### 🖥️ Hardware and Software Setup (please complete the following information)

```shell
not applicable
```

sanzoghenzo:
Hi @kirsch33, thanks for pointing this out.
I've checked the history of that module and it seems it was always like that, maybe @eirannejad could shed some light on that.

But I agree with you, the name `log_errors` suggests that all the errors should/shouldn't  be written.

We have some options here:

- if this is a bug, add the `self._logerror` check before the lines you pointed
- if this was an intended feature, add a `log_commit_errors` argument to the init and use that to suppress the logging, so that we keep retro-compatibility.

kirsch33:
hey @sanzoghenzo 

i am fine with either option and dont mind to submit a PR to address but would like some direction from another maintainer before doing so.

kirsch33:
This is actually causing me some issues now so I've chosen your first option above and created a PR, open to feedback:

[...]

PR #2309, opened 2024-07-09
transaction.py fixes

kirsch33:
See https://github.com/pyrevitlabs/pyRevit/issues/2300

This PR checks for `log_errors=True` before logging ANY error messages (instead of only context errors).

I also modified the log message for `TransactionGroup` to match the others within this file.


sanzoghenzo:
Thank you @kirsch33! I suppose this is the way to go, but I would like to know what @jmcouffin thinks before merging.

jmcouffin:
That sounds about right.
Thanks @kirsch33

PR #2321, opened 2024-07-29
Update EventHandling.cs

acco-jpitts:
Added flag for isOptional in [RegisterUpdater()](https://apidocs.co/apps/revit/2024/edffd44c-2511-c9ee-f330-5cd77414d0e9.htm) to remove warning when updater not found.

image

This will not remove the warning from models that already show it, but new models will not display this error when opened without the pyRevit updater hook.


jmcouffin:
thanks @acco-jpitts 
🚀

PR #2340, opened 2024-08-14
Fix workflow

sanzoghenzo:
Remind me to never use the web IDE to do merge conflict resolution...

The main workflow was not running because a wrong indentation in the yaml file.

I also replaced all the `develop-4` references (mostly in dependabot) and set them to `develop` instead.


jmcouffin:
> Remind me to never use the web IDE to do merge conflict resolution...

😄 

I think we still need to:
- Remove the build-test workflow
- Align the main.yml one (pyRevitCI) with the build test one

PR #2346, opened 2024-08-14
WIP: use ILRepack for single dll builds

sanzoghenzo:
Hi, this is a draft before I leave for a short vacation...

I tried to add ILRepack task after the build, but it complains about not finding RevitAPI (which should be excluded, but it thows the error anyway).

I used [this package](https://github.com/ravibpatel/ILRepack.Lib.MSBuild.Task/tree/master), but I suppose that the complexity of our folder structure gets in our way...

[Here an article](https://www.meziantou.net/merging-assemblies-using-ilrepack.htm) on the subject as reference.

### DIFF
--- a/dev/pyRevitLabs.PyRevit.Runtime/Directory.Build.targets
+++ b/dev/pyRevitLabs.PyRevit.Runtime/Directory.Build.targets
@@ -41,7 +41,7 @@
         <PackageReference Include="ILRepack.Lib.MSBuild.Task" Version="2.0.34.1" />
     </ItemGroup>
 
-    <Target Name="ILRepack" AfterTargets="Build" Condition="'$(Configuration)' == 'Release'">
+    <Target Name="ILRepack" AfterTargets="AfterBuild" Condition="'$(Configuration)' == 'Release'">
         <ItemGroup>
             <InputAssemblies Include="$(TargetPath)"/>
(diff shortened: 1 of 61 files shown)


sanzoghenzo:
Ok, I got to the point that each pyRevitLabs.PyRevit.Runtime.YYYY.dll contains all the dependencies, but they are around 30MB each!

I suppose it is better to use ILRepack on the other pyRevitLabs projects to embed only the 3rd party dependencies...

sanzoghenzo:
This is as far as I can get with my MSBuild knowledge...
I tried to repack the CLI, but it errors out with a wrong path that I don't know how to change.

Now that I think of it, the CLI is the last of our problems, but this is just a test to see if this could be done.

@dosymep when you can, would you like to give it a try?

dosymep:
@sanzoghenzo 

I think it's better not to solve this problem with ILRepack, our build process is already complicated, and this will add even more problems. It's better to focus on fixing errors in the work in the 2025 version, excluding the problem with dependencies (let it be so for now)

sanzoghenzo:
I was trying to test if ILRepack works as of now, to then decide to drop all the modules altogether and use the official nuget packages instead of our custom one.

But if you think is less complicated to sort out dependencies first, it's fine with me!

PR #2353, opened 2024-08-16
Update bundle.yaml

Thomas84:
Title was renamed from Views to Sheets. Reason: It is just wrong in the bundle.yaml title section. Please check if the frensh translation is correct. Maybe the russian title also needs adjustment from view to sheet


jmcouffin:
Checked russian and french translation
rebased to develop branch, it added other changes that were not on fork branch
thanks @Thomas84

PR #2361, opened 2024-08-17
Added some german translations part2

Thomas84:
That's it for today. So many tools, is is unbelievable! But I have improved the German help tooltip a lot, so my users at work don't ask too many questions :P Sorry for the many commits, but I don't know how to do it correctly by using the online GitHub Editor. 

I need to continue at Modify.panel/edit3.stack/Edit.pulldown/Change Element Type Safely.pushbutton/bundle.yaml


jmcouffin:
Danke schön @Thomas84

PR #2362, opened 2024-08-18
Added some german translations part 3

Thomas84:
I think I have finished the German translation of all pyRevit Tools 🤩


jmcouffin:
Super geil!
(Now you know when I learnt German)

PR #2380, opened 2024-08-27
Obsolete IntegerValue property in Revit2024 and higher - Replaced by Value

jmcouffin:
replaced by Value

https://discourse.pyrevitlabs.io/t/release-date-pyrevit-2025/3083/60?u=jean-marc

@acco-jpitts

### DIFF
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Analysis.panel/ColorSplasher.pushbutton/script.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Analysis.panel/ColorSplasher.pushbutton/script.py
@@ -119,7 +119,7 @@ class applyColors(IExternalEventHandler):
                 t.Start()
                 sel_cat = wndw._categories.SelectedItem['Value']
                 value_func = get_value_func()
-                if value_func(sel_cat._cat.Id) == int(BuiltInCategory.OST_Rooms) or value_func(sel_cat._cat.Id) == int(BuiltInCategory.OST_MEPSpaces) or value_func(sel_cat._cat.Id) == int(BuiltInCategory.OST_Areas):
+                if value_func(sel_cat._cat.Id) in (int(BuiltInCategory.OST_Rooms), int(BuiltInCategory.OST_MEPSpaces), int(BuiltInCategory.OST_Areas)):
                     #In case of rooms, spaces and areas. Check Color scheme is applied and if not
                     if version > 2021:
                         if str(wndw.crt_view.GetColorFillSchemeId(sel_cat._cat.Id)) == "-1":
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Analysis.panel/ColorSplasher.pushbutton/script.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Analysis.panel/ColorSplasher.pushbutton/script.py
@@ -47,6 +47,8 @@ from System import Array
 from System.Collections.Generic import *
 
 from pyrevit import HOST_APP, revit
+from pyrevit.compat import get_value_func
+
 
 class subscribeView(IExternalEventHandler):
     def __init__(self):
@@ -116,14 +118,15 @@ class applyColors(IExternalEventHandler):
                 t = Transaction(new_doc, "Apply colors to elements")
                 t.Start()
                 sel_cat = wndw._categories.SelectedItem['Value']
-                if sel_cat._cat.Id.IntegerValue == int(BuiltInCategory.OST_Rooms) or sel_cat._cat.Id.IntegerValue == int(BuiltInCategory.OST_MEPSpaces) or sel_cat._cat.Id.IntegerValue == int(BuiltInCategory.OST_Areas):
+                value_func = get_value_func()
+                if value_func(sel_cat._cat.Id) == int(BuiltInCategory.OST_Rooms) or value_func(sel_cat._cat.Id) == int(BuiltInCategory.OST_MEPSpaces) or value_func(sel_cat._cat.Id) == int(BuiltInCategory.OST_Areas):
                     #In case of rooms, spaces and areas. Check Color scheme is applied and if not
                     if version > 2021:
                         if str(wndw.crt_view.GetColorFillSchemeId(sel_cat._cat.Id)) == "-1":
                             fColorScheme = FilteredElementCollector(new_doc).OfClass(ColorFillScheme).ToElements()
                             if len(fColorScheme) > 0:
                                 for sch in fColorScheme:
-                                    if sch.CategoryId.IntegerValue == sel_cat._cat.Id.IntegerValue:
+                                    if sch.CategoryId == sel_cat._cat.Id:
                                         if len(sch.GetEntries()) > 0:
                                             wndw.crt_view.SetColorFillSchemeId(sel_cat._cat.Id, sch.Id)
                                             break
@@ -228,11 +231,12 @@ class createLegend(IExternalEventHandler):
                 old_all_ele = FilteredElementCollector(new_doc, legends[0].Id).ToElements()
                 ele_id_type = ElementId(0)
                 for ele in old_all_ele:
-                    if ele.Id.IntegerValue != newLegend.Id.IntegerValue and ele.Category != None:
+                    if ele.Id != newLegend.Id and ele.Category != None:
                         if isinstance(ele, TextNote):
                             ele_id_type = ele.GetTypeId()
                             break
-                if ele_id_type.IntegerValue == 0:
+                value_func = get_value_func()
+                if value_func(ele_id_type) == 0:
                      #Get any text in model
                     all_text_notes = FilteredElementCollector(new_doc).OfClass(TextNoteType).ToElements()
                     for ele in all_text_notes:
@@ -479,7 +483,8 @@ class categ_info():
     def __init__(self, cat, param):
         self._name = strip_accents(cat.Name)
         self._cat = cat
-        self._intId = cat.Id.IntegerValue
+        value_func = get_value_func()
+        self._intId = value_func(cat.Id)
         self._par = param
 
 def getActiveView(ac_doc):
@@ -509,7 +514,8 @@ def getValuePar(para):
             value = para.AsValueString()
         elif para.StorageType == StorageType.ElementId:
             id_val = para.AsElementId()
-            if id_val.IntegerValue >= 0:
+            value_func = get_value_func()
+            if value_func(id_val) >= 0:
                 value = Element.Name.__get__(doc.GetElement(id_val))
             else:
                 value ="None"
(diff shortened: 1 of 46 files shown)


sanzoghenzo:
Can I offer a suggestion to avoid code repetition?

```python
value_func  = lambda x: x.Value if HOST_APP.is_newer_than(2023) else x.IntegerValue
```

or a more verbose, but maybe clearer, function to put in `compat.py`

```python
def get_value_func():
    def get_value_2024(item):
        return item.Value

    def get_value_2003(item):
        return item.IntegerValue

    return get_value_2024 if HOST_APP.is_newer_than(2023) else get_value_2003
```

so that one can do, for example

```python
value_func = get_value_func()
sheet_revids = {value_func(x) for x in self.revit_sheet.GetAllRevisionIds()}
add_sheet_revids = {value_func(x) x in self.revit_sheet.GetAdditionalRevisionIds()}
```

pyrevitlib/rpw/db/family.py
```diff
@@ -93,7 +94,9 @@ def in_assembly(self):
         Returns:
             (bool): True if element is inside an AssemblyInstance
         """
-        if self._revit_object.AssemblyInstanceId.IntegerValue == -1:
+        value_func = get_value_func()
+        assembly_id_value = value_func(self._revit_object.AssemblyInstanceId)
+        if assembly_id_value == -1:
             return False
         else:
             return True
```
sanzoghenzo: we can just `return assembly_id_value != -1`, or even `return value_func(self._revit_object.AssemblyInstanceId) != -1`

sanzoghenzo (commented):
This is not properly in-scope, but while you're at it.... 😉

pyrevitlib/rpw/db/family.py
jmcouffin: I totally agree with you:
- not in scope 😄 
- while I am at it.

I fixed a few things here and there but it feels like a never ending story to dig further for now
I thought also about adding docstrings as well ;p

sanzoghenzo:
Just a question: why there were lines like `if cat.Id.IntegerValue == cat_id.IntegerValue`?

I'm not a Revit API guru, but I expect that `cat.Id == cat_id` would do just fine and save some CPU time in the python side of things...

Obviously I don't have the full picture, and I didn't check it, but I can see the [Equals operator for ElementId in revitapidocs](https://www.revitapidocs.com/2023/34a21be8-c836-0ac0-fdc8-d45b112ac580.htm)

jmcouffin:
@sanzoghenzo would you mind having a quick look.
My eyes are sore editing 46 files with dumb changes like this.
I had to stop looking and just do the thing as there are so many coding styles and bad practices in so many tools 🤣

jmcouffin:
I may have broken a thing or two. Time will tell.
A common pattern that bothers me is the ElementId comparison to -1 instead of
```python
if isinstance(element, DB.ElementId.InvalidElementId):
```

extensions/pyRevitTools.extension/pyRevit.tab/Analysis.panel/ColorSplasher.pushbutton/script.py
```diff
@@ -116,14 +118,15 @@ def Execute(self, uiapp):
                 t = Transaction(new_doc, "Apply colors to elements")
                 t.Start()
                 sel_cat = wndw._categories.SelectedItem['Value']
-                if sel_cat._cat.Id.IntegerValue == int(BuiltInCategory.OST_Rooms) or sel_cat._cat.Id.IntegerValue == int(BuiltInCategory.OST_MEPSpaces) or sel_cat._cat.Id.IntegerValue == int(BuiltInCategory.OST_Areas):
+                value_func = get_value_func()
+                if value_func(sel_cat._cat.Id) == int(BuiltInCategory.OST_Rooms) or value_func(sel_cat._cat.Id) == int(BuiltInCategory.OST_MEPSpaces) or value_func(sel_cat._cat.Id) == int(BuiltInCategory.OST_Areas):
```
sanzoghenzo: `if value_func(sel_cat._cat.Id) in (int(BuiltInCategory.OST_Rooms), int(BuiltInCategory.OST_MEPSpaces), int(BuiltInCategory.OST_Areas)):`

extensions/pyRevitTools.extension/pyRevit.tab/Selection.panel/select.stack/Select.pulldown/Find And Select Entities Without Tags.pushbutton/script.py
```diff
@@ -240,14 +242,14 @@
                     tagged_elements.append(List[DB.ElementId](elt.GetTaggedLocalElementIds())[0].Value)
             elif HOST_APP.is_newer_than(2022, or_equal=True):
                 if elt.GetTaggedLocalElementIds() != DB.ElementId.InvalidElementId:
-                    tagged_elements.append(List[DB.ElementId](elt.GetTaggedLocalElementIds())[0].IntegerValue)
+                    tagged_elements.append(List[DB.ElementId](value_func(elt.GetTaggedLocalElementIds())[0]))
```
sanzoghenzo, line 242: Here we can drop the first if block (lines 240-242)

extensions/pyRevitTools.extension/pyRevit.tab/Selection.panel/select.stack/Select.pulldown/Find And Select Entities Without Tags.pushbutton/script.py
```diff
@@ -240,14 +242,14 @@
...
             else:
                 if elt.TaggedLocalElementId != DB.ElementId.InvalidElementId:
-                    tagged_elements.append(elt.TaggedLocalElementId.IntegerValue)
+                    tagged_elements.append(value_func(elt.TaggedLocalElementId))
```
sanzoghenzo: this block is only for revit < 2022, so `IntegerValue` is ok here

pyrevitlib/pyrevit/revit/db/pickling.py
```diff
@@ -38,9 +39,9 @@ class NoneSerializer(Serializable):
 
 class ElementId(Serializable):
     api_types = DB.ElementId
-
+    value_func = get_value_func()
```
sanzoghenzo: move this inside the `__init__` method

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -798,7 +798,7 @@ def get_category(cat_name_or_builtin, doc=None):
                 return cat
     elif isinstance(cat_name_or_builtin, DB.BuiltInCategory):
         for cat in all_cats:
-            if cat.Id.IntegerValue == int(cat_name_or_builtin):
+            if cat.Id == cat_name_or_builtin:
```
sanzoghenzo: `BuiltInCategory` is an enum and not an `ElementId`, so here it should be `value_func(cat.Id) == int(cat_name_or_builtin)`

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -810,29 +810,29 @@ def get_builtincategory(cat_name_or_id, doc=None):
...
         if cat:
-            cat_id = cat.Id.IntegerValue
+            cat_id = cat.Id
     elif isinstance(cat_name_or_id, DB.ElementId):
-        cat_id = cat_name_or_id.IntegerValue
+        cat_id = cat_name_or_id
     if cat_id:
         for bicat in DB.BuiltInCategory.GetValues(DB.BuiltInCategory):
-            if int(bicat) == cat_id:
+            if bicat == cat_id:
```
sanzoghenzo: same as above, `int(bicat) == value_func(cat_id)`

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -849,9 +849,12 @@ def get_builtinparameter(element, param_name, doc=None):
     doc = doc or DOCS.doc
     eparam = element.LookupParameter(param_name)
     if eparam:
+        eparam_def_id = eparam.Definition.Id
         for biparam in DB.BuiltInParameter.GetValues(DB.BuiltInParameter):
-            if int(biparam) == eparam.Definition.Id.IntegerValue:
+            if biparam == eparam_def_id:
```
sanzoghenzo: same as above, biparm is an enum

pyrevitlib/rpw/db/builtins.py
```diff
@@ -155,8 +156,10 @@ def from_category_id(self, category_id):
         Returns:
             ``DB.BuiltInCategory`` member
         """
-        bic = Enum.ToObject(DB.BuiltInCategory, category_id.IntegerValue)
-        if DB.ElementId(bic).IntegerValue < -1:
+        value_func = get_value_func()
+        bic = Enum.ToObject(DB.BuiltInCategory, value_func(category_id))
+        bic_value = value_func(DB.ElementId(bic))
```
sanzoghenzo: To avoid back and forth between ElementIds, integers and enums, I would do like this (untested)

```python
cat_id_value = value_func(category_id)
if cat_id_value < -1 and Enum.IsDefined(DB.BuiltInCategory, cat_id_value):
    return Enum.ToObject(DB.BuiltInCategory, cat_id_value)
```

pyrevitlib/rpw/db/family.py
```diff
@@ -93,10 +94,8 @@ def in_assembly(self):
         Returns:
             (bool): True if element is inside an AssemblyInstance
         """
-        if self._revit_object.AssemblyInstanceId.IntegerValue == -1:
-            return False
-        else:
-            return True
+        value_func = get_value_func()
+        return value_func(self._revit_object.AssemblyInstanceId) != -1
```
sanzoghenzo: `return self._revit_object.AssemblyInstanceId != DB.ElementId.InvalidElementId`

sanzoghenzo (changes requested):
You did a hell of a job!

some of the variables are enum items, so we should use the value to compare them.
I suggested only one place in which to use DB.ElementId.InvalidElementId (it's not a type so we should use comparison operartors and not `isinstance`), but there may be other places it can be used.

extensions/pyRevitTools.extension/pyRevit.tab/Analysis.panel/ColorSplasher.pushbutton/script.py
jmcouffin: it makes sense, thks

sanzoghenzo:
Argh, I was going to suggest you to do a merge with squash to avoid all those commits in the git log 😅 

It's getting harder and harder browse the commit history with all those single file commits....

sanzoghenzo:
Did you test this PR? It is broken since you used `__revit__.Application.is_newer_than`, but the Revit Application class doesn't have that method. You need to use `HOST_APP.is_newer_than` to make it work.

I'll fix it after lunch

jmcouffin:
> Did you test this PR?
I haven't
I guess HOST_APP won't cut it either as it is initiated in the init, right? Inception situation?
something like `__revit__.Application.VersionNumber > 2023` should work

jmcouffin:
and for the sake of documentation, this release works in 2025 but not in 2024 and below
image

sanzoghenzo:
That error is the one I'm trying to fix on my PR 😉 

I'll factor out the version number retrieval from the `is_netcore` to use it for this function, too.

jmcouffin:
> I'll fix it after lunch

You were not kidding! @sanzoghenzo Nice!

PR #2414, opened 2024-10-11
Create radar_check.py

tay0thman:
This Model Checker swiftly verifies the extents of the Revit model.  Placing model extents more than 10 miles from the project's  internal origin can lead to issues with accuracy, tolerance,  performance, and viewport display. This check ensures that the  model remains within a 10-mile radius of the internal origin.


jmcouffin:
Nice, I will have a look tomorrow

jmcouffin (approved):

dosymep:
@jmcouffin 
I just scheduled a review to check tomorrow :D

PR #2418, opened 2024-10-16
Update radar_check.py - Rev1

tay0thman:
Hello @dosymep 

Made the changes below per @jmcouffin feedback.

_1- you may want to present digits results in a tabular form for readability and ease in comparison._

**Response: Added output.print_table**

_2- the project base point cannot be placed at more than 10 miles in more recent version of Revit_

**Response: Good to know! I removed the report for the location of the project basept.**

_3- you may want to use the convert unit util to display the distances in the same units as in the UI_

**Response: I used the unitutils to convert the internal distances to the model distances, however in metric I am having hard time to show the suffix (mm, M, km)....etc**

_4- you may want to limit the amount of objects output in the output windows using linkify as it may greatly slow the tool_

**Response: Set the limit of linkify to 10 prints per category, if there are more violating objects, then it will show a message "Manual investigation is required"**

_5- I tend to prefer print_md over print_html as it is more efficient and improves readability_

**Response: I switched most of them to use print_md. also, I used pyrevit emojis instead of html colors.**

Please look and let me know if this makes sense.

image

When violations are detected:
image

### DIFF
--- a/extensions/pyRevitTools.extension/checks/radar_check.py
+++ b/extensions/pyRevitTools.extension/checks/radar_check.py
@@ -18,15 +18,14 @@ clr.AddReference('RevitAPI')
 from Autodesk.Revit.DB import *
 
 #pyRevit
-from pyrevit import revit, DB, DOCS
+from pyrevit import DB, DOCS, HOST_APP
 from pyrevit import script
-from pyrevit import forms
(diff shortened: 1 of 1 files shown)


dosymep:
@jmcouffin code have some problems `the developer forgot to write a method call`, and some other little things

image

PR #2423, opened 2024-10-18
Create CADAUDIT_check.py

nasmovk:
My proposal for a comprehensive audit of cad instances. See details described in the ModelChecker class comments. It would replace the old LISTDWGs by F. Beaupere, from the Project panel, link tools. Bonus: It also extends the basic functionality of RPW by adding radio buttons and a cancel button.


jmcouffin:
> It also extends the basic functionality of RPW by adding radio buttons and a cancel button.

would be cool if you could add it to the rpw module and refer to it

dosymep:
please add screenshots or any results how to work your script

nasmovk:
@dosymep Sure.
The check starts like this: 
image
Then you get a choice of run mode (using RPW with added radio buttons)
image
The output is in a table with clickable links back to the elements and a legend of explanations of the checks made and why.
image
There is some feedback on it in the forum:
https://discourse.pyrevitlabs.io/t/hackathon-preflight-checks-maker-fair/7702/8

nasmovk:
> > It also extends the basic functionality of RPW by adding radio buttons and a cancel button.
> 
> would be cool if you could add it to the rpw module and refer to it

Yes. I thought about doing that, but I didn't think it was allowed to change the built-in libraries. How would I do it? Make a PR on the RPW files in question?

jmcouffin:
@nasmovk yes, that and writing and adding it to the rpw test in the dev extension.

The rpw lib has been maintained mostly here by pyrevit users

extensions/pyRevitTools.extension/checks/CADAUDIT_check.py
```diff
@@ -0,0 +1,200 @@
...
+    #cadinstances = DB.FilteredElementCollector(doc).OfClass(DB.ImportInstance).WhereElementIsNotElementType().ToElements()
+    if active_view_only:
+        cadinstances = DB.FilteredElementCollector(doc, ac_view.Id).OfClass(DB.ImportInstance).WhereElementIsNotElementType().ToElements()
+    else:
+        cadinstances = DB.FilteredElementCollector(doc).OfClass(DB.ImportInstance).WhereElementIsNotElementType().ToElements()
+    
+    if len(cadinstances) >0:
+        return cadinstances, len(cadinstances)
+    else:
+        print("No CAD instances found in the {}.".format("active view" if active_view_only else "model"))
```
dosymep: perhaps it is better to use [forms.alert(exitscript=True)](https://docs.pyrevitlabs.io/reference/pyrevit/forms/#pyrevit.forms.alert) here

image

extensions/pyRevitTools.extension/checks/CADAUDIT_check.py
```diff
@@ -0,0 +1,200 @@
...
+
+    return user_input_dict
+
+def get_load_stat(cad, is_link):
+    """ Loaded status from the import instance's CADLinkType """
+    cad_type = doc.GetElement(cad.GetTypeId()) # Retreive the type from the instance
+    
+    if is_link:
+        exfs = cad_type.GetExternalFileReference()
+        status = exfs.GetLinkedFileStatus().ToString()
```
dosymep: to compare enumerations, it is better not to convert them to string `if status == LinkedFileStatus.Loaded`, [see docs](https://www.revitapidocs.com/2022/410b48f8-2a32-3d54-492c-c9a9fe4030ca.htm)

dosymep (changes requested):

extensions/pyRevitTools.extension/checks/CADAUDIT_check.py
nasmovk: I was getting the string value to be able to return it as a message

extensions/pyRevitTools.extension/checks/CADAUDIT_check.py
jmcouffin: fixed by commit 9fd9c57950dc859e12dd2c45bf724827d94fc838

extensions/pyRevitTools.extension/checks/CADAUDIT_check.py
jmcouffin: agree with @dosymep, not the most elegant way/
will merge anyway for now

PR #2425, opened 2024-10-18
Update radar_check.py - Code Cleanup, expanding tabulated info

tay0thman:
Updated the script for better compliance and fixed issues reported by @jmcouffin and @dosymep
expanded the table to show more breakdown of information, however fractional display is still a question.

image


jmcouffin:
> however fractional display is still a question

AsValueString() should get you the values as displayed in the UI

tay0thman:
I'll give this a try! thank you for reviewing it again.

Issue #2352, opened 2024-08-16
Balloon forms in lieu of toast (due to pyRevit 5)


aussieBIMguru:
Looks like the toast got burnt in pyRevit5!

Just leaving this here if it helps. I noticed in pyRevit 5 toast forms do not seem to trigger. From what I understand of the toast form it depends on another code library packaged into pyRevit that I suspect isn't netcore compatible.

I found the windows balloonforms work and can generate a pretty similar outcome, maybe below code might help the team if the toast form is difficult to reimplement. The main drawback is it's a bit less pretty and doesn't support the dictionary list for buttons like toast did.

The on click function is a bit rudimentary that I've built - attempts to support filepath/urlpath open on clicking the form generally vs giving the user extra buttons with tooltip to describe this being the behavior of doing so.

```
import clr
import os
import webbrowser
clr.AddReference('System.Windows.Forms')
clr.AddReference('System.Drawing')
import System
from System.Windows.Forms import Application, ToolTipIcon, NotifyIcon

class formUtils_balloonMessage:
    def __init__(self, title, msg, warnIcon = False, filePath = None, urlPath = None):
        self.title = title
        self.msg = msg
        self.filePath = filePath
        self.urlPath = urlPath
        if warnIcon:
            self.icon = ToolTipIcon.Warning
        else:
            self.icon = ToolTipIcon.Info
    def show(self):
        balloonMessage = NotifyIcon()
        if self.filePath is not None or self.urlPath is not None:
            balloonMessage.BalloonTipClicked += self.onClick
        balloonMessage.Text = self.title
        balloonMessage.Visible = True
        balloonMessage.Icon = System.Drawing.Icon.ExtractAssociatedIcon(Application.ExecutablePath)
        balloonMessage.ShowBalloonTip(5000, self.title, self.msg, self.icon)
    def onClick(self, sender, args):
        if self.filePath is not None:
            try:
                os.startfile(self.filePath)
            except:
                pass
        elif self.urlPath is not None:
            try:
                webbrowser.open(self.urlPath)
            except:
                pass
```

aussieBIMguru:
Example of the appearance vs toast.

![image](https://github.com/user-attachments/assets/153c4aae-6bb6-4855-adac-6af32bc61a6a)

jmcouffin:
Is it possible this is not a bug @aussieBIMguru 
Toast notification can be deactivated in Windows (at least this is what I have done on my setup long ago)
[...]

PR #2426, opened 2024-10-19
fix: make toasts work again

sanzoghenzo:
this makes toasts work again in revit 2025.

The issue was mentioned in #2352 

The problem was the `pyrevit-toast.exe` moved to the `netfx` folder, but since it is a compiled go binary, and only used by the toaster.py module, I moved it into the pyrevit\forms package.

The "Test Toast" in the pyRevitDev Unit tests section works with both revit 2023 and 2025 versions on my machine now.


dosymep (approved):
good work 👍

jmcouffin (approved):
You manage to 'make toast great again'
Was it a ref to the former USA a.hole president motto? 😁

Issue #2429, opened 2024-10-23
[Bug]: update button in 2025


jmcouffin:
### ✈ Pre-Flight checks

- [X] I **don't have SentinelOne** antivirus installed (see above for the solution)
- [X] I have **searched in the issues** (open and closed) but couldn't find a similar issue
- [X] I have **searched in the [pyRevit Forum](https://discourse.pyrevitlabs.io)** for similar issues
- [X] I already **followed the [installation troubleshooting guide](https://pyrevitlabs.notion.site/Installation-issues-f6495adb56254e9e8021f8a68e3b2ab7) thoroughly**
- [X] I am using the **latest pyRevit Version**

### 🐞 Describe the bug

when running the update button in revit 2025.3, it throws the following error:
`INFO [pyrevit.versionmgr.updater] Checking internet connection... `

### ⌨ Error/Debug Message

```shell
INFO [pyrevit.versionmgr.updater] Checking internet connection...
```

### ⏲️ Expected behavior

to update extensions

### 🖥️ Hardware and Software Setup (please complete the following information)

```shell
-
```

### Additional context

The checking of the internet connection is not working in 2025, I need to check later

jmcouffin:
ref: #2412

PR #2430, opened 2024-10-23
Update __init__.py

jmcouffin:
fix for #2429

I used the requests module to get it to work in all revit version


jmcouffin:
I will need to have a closer look at it, I introduced a bug doing this

PR #2432, opened 2024-10-24
Fix #2429 - Update button compat with NETCORE

### LINKED ISSUE #2429: [Bug]: update button in 2025

jmcouffin:
fix #2429

### DIFF
--- a/.pyrevitargs
+++ b/.pyrevitargs
@@ -1,0 +1,3 @@
+https://github.com/pyrevitlabs/pyRevit/archive/master.zip
+master
+basepublic
--- a/.pyrevitargs
+++ b/.pyrevitargs
@@ -1,3 +1,0 @@
-https://github.com/pyrevitlabs/pyRevit/archive/master.zip
-master
-basepublic
--- a/pyrevitlib/pyrevit/coreutils/__init__.py
+++ b/pyrevitlib/pyrevit/coreutils/__init__.py
@@ -24,12 +24,10 @@ from collections import defaultdict
 
 #pylint: disable=E0401
 from pyrevit import HOST_APP, PyRevitException
-from pyrevit import compat
-from pyrevit.compat import PY3, PY2, NETCORE
+from pyrevit.compat import PY3, PY2
 from pyrevit.compat import safe_strtype
 from pyrevit.compat import winreg as wr
 from pyrevit import framework
-from pyrevit import api
 
 # RE: https://github.com/pyrevitlabs/pyRevit/issues/413
 # import uuid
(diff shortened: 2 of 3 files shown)


dosymep:
@jmcouffin maybe try to use [Ping](https://learn.microsoft.com/ru-ru/dotnet/api/system.net.networkinformation.ping?view=netframework-4.8)?

jmcouffin:
> @jmcouffin maybe try to use [Ping](https://learn.microsoft.com/en-us/dotnet/api/system.net.networkinformation.ping?view=netframework-4.8)?

I will have a look eventually, 
I am a bit swamped with work right now.

sanzoghenzo:
Do we really need to check the connection before doing anything?

- `can_access_url` is only used by `check_internet_connection`;
- `check_internet_connection` uses (not-so-)random urls that have nothing to do with the update process (well, only github.com does), and returns the first successful url, but nobody cares about it
- `check_internet_connection` is only used by `versionmgr.updater._check_connection` that actually "cares" about that url, just to display a debug message and return True

[This SO answer](https://stackoverflow.com/a/33117579/4149512) suggests using socket and a fixed ip address and tcp port instead of relying on name resolutions and application layers.
Here's the code, that I would put in the _check_connection function and discard the other two functions
Not sure if it plays well with IronPython/python.net, though...

```python
import socket

def _check_connection(host="8.8.8.8", port=53, timeout=3):
    """
    Host: 8.8.8.8 (google-public-dns-a.google.com)
    OpenPort: 53/tcp
    Service: domain (DNS/TCP)
    """
    try:
        socket.setdefaulttimeout(timeout)
        socket.socket(socket.AF_INET, socket.SOCK_STREAM).connect((host, port))
        return True
    except socket.error as ex:
        print(ex)
        return False
```

sanzoghenzo:
I didn't test it, but something like this should work

```python
from System.Net.NetworkInformation import Ping, IPStatus

def _check_connection(ip_address="8.8.8.8"):
    ping_sender = Ping()
    try:
        reply = ping_sender.Send(ip_address)
    except Exception:
        return False
    return reply.Status == IPStatus.Success
```

jmcouffin:
> I didn't test it, but something like this should work
> 
> ```python
> from System.Net.NetworkInformation import Ping, IPStatus
> 
> def _check_connection(ip_address="8.8.8.8"):
>     ping_sender = Ping()
>     try:
>         reply = ping_sender.Send(ip_address)
>     except Exception:
>         return False
>     return reply.Status == IPStatus.Success
> ```

this did not work.

jmcouffin:
> ~Do we really need to check the connection before doing anything?~ It makes sense if we don't want to crash the code with timeout exceptions... anyway:
> 
> * `can_access_url` is only used by `check_internet_connection`;
> * `check_internet_connection` uses (not-so-)random urls that have nothing to do with the update process (well, only github.com does), and returns the first successful url, but nobody cares about it
> * `check_internet_connection` is only used by `versionmgr.updater._check_connection` that actually "cares" about that url, just to display a debug message and return True
> 
> [This SO answer](https://stackoverflow.com/a/33117579/4149512) suggests using socket and a fixed ip address and tcp port instead of relying on name resolutions and application layers. Here's the code, that I would put in the _check_connection function and discard the other two functions Not sure if it plays well with IronPython/python.net, though...
> 
> ```python
> import socket
> 
> def _check_connection(host="8.8.8.8", port=53, timeout=3):
>     """
>     Host: 8.8.8.8 (google-public-dns-a.google.com)
>     OpenPort: 53/tcp
>     Service: domain (DNS/TCP)
>     """
>     try:
>         socket.setdefaulttimeout(timeout)
>         socket.socket(socket.AF_INET, socket.SOCK_STREAM).connect((host, port))
>         return True
>     except socket.error as ex:
>         print(ex)  # change it with some logging
>         return False
> ```
> 
> Obviously the .net Ping solution could be the best solution, so that we can port the core of pyrevit to .NET more easily 😉

on the contrary, this did work (tested in 2022 and 2025)

tay0thman:
Testing WIP build: [5.0.0.24309+2043-wip]

IronPython Traceback:
Traceback (most recent call last):
  File "C:\Program Files\pyRevit-Master\bin\netcore\engines\pyRevitLoader.py", line 30, in <module>
  File "C:\Program Files\pyRevit-Master\pyrevitlib\pyrevit\loader\sessionmgr.py", line 30, in <module>
  File "C:\Program Files\pyRevit-Master\pyrevitlib\pyrevit\versionmgr\updater.py", line 14, in <module>
**ImportError: No module named NetworkInformation**

PR #2441, opened 2024-10-29
Create Naming_Convention_Check

jmcouffin:
From @andreasdraxl 

> This code is based on a true story.
> 
> Naming conventions are the foundation of a project. Naming is essential for model checks, filters, and view settings. Until now, I created my checks using Revit export text files with Power-BI. PyRevit enables continuous checking of all categories against their predefined names with just a click.
>


jmcouffin:
@andreasdraxl 
Making the wall type list as an external XML or JSON file would be better 
Even adding a file loader with a default file/template.
This way you could have different QA QC setups

andreasdraxl:
@jmcouffin 
so i found this page...  i created other branche. yes it runs more smooth now and is addaptable for others.
2024-10-30_17h46_26
2024-10-30_17h46_48

jmcouffin:
A few things_ I don't have much time these days, so I will stay high level: 
- you should be able to add commits to your **patch-1** branch and they will appear in this Pull request. Or you could edit your file directly in the patch-1 branch with the 'same' result.
- your file name is incorrect: Naming_Convention does not have an extension .py nor the end of the file name like so: check.py - It should be something like wall_types_naming_convention_check.py
- you should add a file selector for the UI, as today it points to your Documents file path. Use the [pyrevit.forms ](https://docs.pyrevitlabs.io/reference/pyrevit/forms/#pyrevit.forms.pick_file) 
- add your sample wall types file in json in this folder.


Consider your entry valid @andreasdraxl

andreasdraxl:
i will add pyRevit forms , than it will be selectable
2024-10-30_19h41_53

jmcouffin:
you may want to set the default path to point to the checks folder, so that it finds your sample right away

andreasdraxl:
2024-10-30_19h59_34
ok...  thats more comfort

jmcouffin:
- Better to use pick_file from the pyrevit.forms module
- point to the checks folder of the pyrevit install folder `extensions/pyRevitTools.extension/checks` and put the walls types list there

andreasdraxl:
i will do it, i think for my project, i think for hackathon - work is done.

jmcouffin:
@andreasdraxl can you check your branch and patch1 fork, i made a few PR to fix remaining stuffs. Please merge the PR accordingly

PR #2455, opened 2024-11-10
Update updater.py

jmcouffin:
fixing internet connection check using socket


jmcouffin:
fix for #2429 and regression #2432

PR #2467, opened 2024-11-27
Update radar_check.py

tay0thman:
Cleaned up functions,
Removed redundancy,
streamlined tabulation of coordinates, distances, and angles. used regions to divide functions.
other formatting items


jmcouffin:
Good job @tay0thman

PR #2478, opened 2024-12-09
Update script.py -- Color Splasher 'Reset All Colors' fix

utilitas-firmitas:
'Reset All Colors' button wouldn't work in Revit (colors remained).  This updates the `ResetColors` class to make the action work.

Updated the following:
1. Transaction management using context managers (`with` statements)
2. Separate transactions for filter removal and override reset
3. Added logic to handle both selective (when category selected) and global reset
4. Better error handling and exception reporting
5. Added null checks for view and document
6. More efficient element collection and filtering
7. Cleanup of override graphics settings


jmcouffin:
Thanks @utilitas-firmitas

PR #2489, opened 2024-12-26
docs: add developer guides

sanzoghenzo:
I finally found the time to create what I promised:

- turned the developer guide that we discussed on the forum to a docs page
- added an architecture guide to understand the loading process and script execution
- added a page with a really brief description of the various folders found in the repository

I updated MkDocs and dependencies in the process, so if you want to check it out locally you need to

```shell
pipenv sync
pipenv run mkdocs serve
```

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -24,14 +24,6 @@
             "markers": "python_version >= '3.8'",
             "version": "==2.16.0"
         },
-        "backports-strenum": {
-            "hashes": [
-                "sha256:77c52407342898497714f0596e86188bb7084f89063226f4ba66863482f42414",
-                "sha256:cdcfe36dc897e2615dc793b7d3097f54d359918fc448754a517e6f23044ccf83"
-            ],
-            "markers": "python_version < '3.11'",
-            "version": "==1.3.1"
-        },
         "black": {
             "hashes": [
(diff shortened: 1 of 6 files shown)


jmcouffin (approved):

jmcouffin:
This great @sanzoghenzo 
And very much needed for the future of the beast

jmcouffin:
As if it were that simple 🤣😁😐
https://github.com/pyrevitlabs/pyRevit/actions/runs/12508207664/job/34895953611#step:10:1

sanzoghenzo:
maybe we can solve that by changing [this line](https://github.com/pyrevitlabs/pyRevit/blob/cab55a3c7365a1c8aa96529b46b29340a350729f/.github/workflows/docs.yml#L38) to `pipenv sync` (this will make sure that the packages used are the same as in the pipfile.lock, so exactly the same as the ones I used on my machine).

I'll try tomorrow, I'm too tired now to break things now :sweat_smile:

jmcouffin:
Trying it right now.
Let's break things a bit more.

jmcouffin:
Neither broken nor fixed

sanzoghenzo:
Oh, I now see that colorama is flagged to be installed only on windows

https://github.com/pyrevitlabs/pyRevit/blob/9fbe2d3bbf10c590cc470980c3925b21e1e963f9/Pipfile.lock#L183

I'll try to fix it as soon as I get back to my pc

PR #2491, opened 2024-12-27
build: ensure colorama is installed in CI

sanzoghenzo:
This should fix the error we got for the docs CI.


jmcouffin (approved):

jmcouffin:
Alive and well.
Great job again @sanzoghenzo

PR #2493, opened 2024-12-27
fix: show unicode in toasts

### LINKED ISSUE #2484: [Bug]: forms.toast() does not support Unicode characters. (Enhancement)
### ✈ Pre-Flight checks

### 🐞 Describe the bug

All text fields of function **forms.toast()** do not support Unicode characters.
Please add support.
toast

### ⌨ Error/Debug Message

```shell
No errors.
Just scribbles.
```

### ♻️ To Reproduce

_No response_

### ⏲️ Expected behavior

_No response_

```shell
Win10.
Revit 2023.1.6
5.0.0.24325+1012-wip:2712:2023.1.60
```

### Additional context

_No response_

sanzoghenzo:
closes #2484

Title and message with unicode chars (I tried emojis) work, actions and sounds don't.

Now that I think of it, we could use [the `Microsoft.Toolkit.Uwp.Notifications` .NET package](https://learn.microsoft.com/en-us/windows/apps/design/shell/tiles-and-notifications/send-local-toast?tabs=desktop) for that instead of relying on a powershell script...


dosymep:
I think it's a bad idea to use a new library with a thousand obscure dependencies from another framework (not wpf)

sanzoghenzo:
I should have tried it inside pyRevit before submitting it 😅 I simplified some things and made the function signature compatible with `pyrevit.forms.toast`, so that I could drop it (I still import the `toast` function inside forms `__init__`, so nothing changed for the user).

The problem now is... IronPython's ElementTree doesn't seem to work with unicode, so I'm back at square one 😓 

@dosymep I saw that you wrote a comment about the dependency, do you think that using an UWP component will break things?

dosymep:
> do you think that using an UWP component will break things?
yep
how it is work? why do you use powershell?

sanzoghenzo:
> > do you think that using an UWP component will break things?
> 
> yep 

Would it be better to use [`Windows.UI.Notifications`](https://learn.microsoft.com/it-it/uwp/api/windows.ui.notifications.toastnotificationmanager?view=winrt-26100)?

> how it is work? why do you use powershell?

Because I just ported the go code of the previous `pyrevit-toaster.exe`, or to be precise a fork that should have solved the unicode issue (but used string templating instead of an xml library to build the configuration).
I just removed a step, I call directly powershell instead of a go binary that in turn calls powershell.
The powershell calls in fact the `Windows.UI.Notifications.ToastNotificationManager` to display the toast.

dosymep:
> go binary that in turn calls powershell
oh okay :)

sanzoghenzo:
I am so close... but I don't know how to solve it!

I switched to string templates to avoid unwanted etree string escaping;
I also fixed the flashing powershell window (i forgot to add a flag to subcommand)

Now the Test Button in DevTools shows a "??" instead of the rocket, but if I copy the script from the debug log and paste into powershell with `powershell -NoProfile -NonInteractive -Command ` prepended, the toast correctly shows the emoji (even thow it is shown as "??" in the powershell window)...

I tried to encode/decode the strings, but no combination works... As always, ChatGPT is full of shit.
Any Ideas how to properly pass the unicode text to powershell as-is?

BTW, I also tried using Windows.UI.Notification, but obviously it needed an assembly that is not referenced, and I'm too lazy to try to add it (where? should it be pyRevitLabs.Runtime?) and recompile 😅

dosymep:
I would help you, but for some reason notifications don't work on my version of the operating system

sanzoghenzo:
I tried to add the patch for windows subprocess, it shows the correct toast but `subprocess.check_output` throws an exception.
I tried various iterations with `Popen`, `call` and `check_call`, they give me a TypeError that I can't understand (something related to the ctypes).

Since this isn't an high priority task, I'm giving up on this and remove the issue form the pyrevit 5 RC milestone, if someone else wants to pick it up it's more than welcome to do so!

PR #2503, opened 2025-01-07
Full models audit

jmcouffin:
Preflight audit of all models, including linked models.

⚠️ Links must be loaded

This QC tools returns the following data:
- project name, number, client, phases, worksets
- element count
- purgeable elements count
- all warnings count
- critical warnings count
- rvtlinks count
- activated analytical model elements count
- rooms count
- unplaced rooms count
- sheets count
- views count
- views not on sheets
- schedules not sheeted count
- copied views count
- view templates count
- unused view templates count
- filters count
- unused view filters count
- materials count
- line patterns count
- dwgs count
- linked dwg count
- inplace family count
- not parametric families count
- family count
- imports subcats count
- generic models types count
- detail components count
- text notes types count
- text notes types with width factor != 1
- text notes count
- text notes with allCaps applied
- detail groups count
- detail groups types count
- model group count
- model group type count
- reference planes count
- unnamed reference planes count
- detail lines count
- dimension types count
- dimension count
- dimension overrides count
- revision clouds count


extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
+# -*- coding: UTF-8 -*-
+# flake8: noqa
```
sanzoghenzo: This completely disables code quality checks, and since we don't have strict policies in places (like pre-commit checks that prevents committing "bad" code), I believe this is useless and harmful (for when we would add such mandatory checks)

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+        else:
+            pass
+    except Exception as e:
+        print(format_exc())
+        print(e)
+        output.self_destruct(30)
+
+
+
+timer = Timer()
```
sanzoghenzo: The timer start when you load the script, not when the TestCase starts... I believe this should be moved inside the `setUp` method of the ModelChecker class as an attribute

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+           "model_group_count", "model_group_type_count",
+           "reference_planes_count", "unnamed_ref_planes_count",
+           "detail_lines_count", "dim_types_count", "dim_count",
+           "dim_overrides_count", "revision_clouds_count"
+           ]
+
+
+def check_model(doc, output):
+    try:
+        if not doc.IsFamilyDocument:
```
sanzoghenzo: since the "else" branch does nothing, invert the condition and return

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+                            # print('Data already in csv file')
+                            break
+                if not flag:
+                    with open(EXPORT_FILE_PATH, mode='ab') as csv_file:
+                        # print('writing')
+                        w = writer(csv_file, lineterminator='\n')
+                        w.writerow(data)
+
+            # RVTLinks
+            if rvtlinks_count != 0:
```
sanzoghenzo: move everything inside this branch to a separate function.
This also prevents overwriting the variables used for the main doc (could it be the issue that you had with the doc_cached_issues?)
And maybe this could be DRYed and used for both main doc an the links...

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+                            groups_summary_frame + \
+                            reference_planes_frame + \
+                            create_frame(
+                                'Materials',
+                                card_builder(100, materials_count, ' Materials')
+                                )
+
+                        output.print_html(html_content + '</div>')
+                    else:
+                        pass
```
sanzoghenzo: useless, remove

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+                    w.writerow(data_str)
+            else:
+                with open(EXPORT_FILE_PATH, mode='rb') as csv_file:
+                    r = reader(csv_file, delimiter=',')
+                    flag = False
+                    for row in r:
+                        if row[1] == date and row[2] == doc_clean_name:
+                            flag = True
+                            # print('Data already in csv file')
+                            break
```
sanzoghenzo:
```suggestion
                    flag = any(row[1] == date and row[2] == doc_clean_name for row in r)
```

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+                if not flag:
+                    with open(EXPORT_FILE_PATH, mode='ab') as csv_file:
+                        # print('writing')
+                        w = writer(csv_file, lineterminator='\n')
+                        w.writerow(data)
```
sanzoghenzo: extract to a separate function.

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+
+            # RVTLinks
+            if rvtlinks_count != 0:
+                output.print_md('# RVTLinks')
+                for rvtlink in rvtlinks_docs:
+                    link_name = doc_name(rvtlink)
+                    link_printed_name = clean_name(rvtlink)
+                    output.print_md('## ' + link_printed_name)
+                    output.print_md('___')
+                    if rvtlink is not None:
```
sanzoghenzo: how can this be None?

anyway, invert the condition and skip early

```suggestion
                    if rvtlink is None:
                        continue

```

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+            # Warnings
+            all_warnings_count, _, warnings_guid = warnings(doc)
+            critical_warnings_count = critical_warnings(warnings_guid, CRITICAL_WARNINGS)
+            if all_warnings_count > 0:
+                try:
+                    show_balloon(header='Warnings', text='The file {} contains {} warnings'.format(doc_clean_name, all_warnings_count), is_favourite=True, is_new=True)
+                except Exception as e:
+                    print(e)
+            # Links
+            _, rvtlinks_elements_items, rvtlinks_count, type_status, rvtlinks_docs = rvtlinks_elements(
```
sanzoghenzo: you use the `rvtlinks_elements` twice in you code and both times you don't need the first item... you might as well avoid returning it in the first place!

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+                    projects_numbers.append(project_number)
+                    projects_clients.append(project_client)
+                    projects_phases.append(phases(doc))
+                    projects_worksets.append(worksets(doc))
+                rvtlinks_docs_name, rvtlinks_instances_name = rvt_links_name(
+                    rvtlinks_elements_items)
+                rvtlinks_unpinned = rvt_links_unpinned_count(
+                    rvtlinks_elements_items)
+                rvtlinks_unpinned_str = rvt_links_unpinned_str(
+                    rvtlinks_elements_items)
```
sanzoghenzo: Move this block to line 180 and you'll find out that you could directly create a single list with the info of a single link in a row, ready for the table!

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+            critical_warnings_count = critical_warnings(warnings_guid, CRITICAL_WARNINGS)
+            if all_warnings_count > 0:
+                try:
+                    show_balloon(header='Warnings', text='The file {} contains {} warnings'.format(doc_clean_name, all_warnings_count), is_favourite=True, is_new=True)
+                except Exception as e:
+                    print(e)
+            # Links
+            _, rvtlinks_elements_items, rvtlinks_count, type_status, rvtlinks_docs = rvtlinks_elements(
+                doc)
+            if rvtlinks_count != 0:
```
sanzoghenzo: the `count` returned by the function is nothing more than `len(items)`, and the pythonic way to check if a list is empty is 
```suggestion
            if rvtlinks_elements_items:

```

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+            body_css = '<style>.grid-container { display: grid; justify-content: center; align-items: center; }</style><div class="grid-container">'
+            
+            # Main file infos
+            project_info = [project_name, project_number,project_client, project_phases, worksets_names, "N/A", "N/A", "N/A", "N/A"]
+            output.print_md('# Main File Infos')
+            output.print_table([project_info], columns=['Project Name', 'Project Number', 'Client Name', 'Project Phases', 'Worksets', 'Linked File Name', 'Instance Name','Loaded Status', 'Pinned status'])
+
+            # Linked files infos
+            links_cards = ''
+            if rvtlinks_count != 0:
```
sanzoghenzo:
```suggestion
            if rvtlinks_elements_items:

```

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
```diff
@@ -0,0 +1,478 @@
...
+            # Dimensions
+            dim_types_count, dim_count, dim_overrides_count = count_dimension_types(doc), count_dimensions(doc), count_dimension_overrides(doc)
+            # Revision clouds
+            revision_clouds_count = revisions_clouds(doc)
+
+            # output section
+            output.set_height(900)
+            output.set_width(1200)
+            output.close_others()
+            output = script.get_output()
```
sanzoghenzo: Why are you rewriting an input variable? is this needed?

sanzoghenzo (changes requested):
Quick first review, I didn't check the query module yet.
Just thinking out loud: I get that you wanted to separate the data collection from the output, but since the logic is already in the query module, wouldn't it be better to group the the data variables to the related frame and then split each group to its own function?

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
jmcouffin: I just added this as my vscode is not properly setup and double checks everything for some reason, removing

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
jmcouffin: you know, just before going to bed, I thought of that one, and that it would be the first thing you tell me ;p

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
jmcouffin: I tried moving it in the setUp method, and it did not provide the right result.
I don't know how I should do that

jmcouffin:
I still need to review the numbers outputed and maybe get to solve why the document is not properly stored / cached

extensions/pyRevitTools.extension/checks/modelchecker_new_check.py
sanzoghenzo: Doesn’t `self.timer = Timer()` work? Obviously you then need to use `self.timer` in the other method (maybe this is what you tried, but this is just to be sure we're aligned 😉)

jmcouffin: https://media.giphy.com/media/Kzu05A116shtnT436d/giphy.gif?cid=790b7611jo8m9lrtc81vlk2ecf7l90ec9gb60zbpfrvw8txk&ep=v1_gifs_search&rid=giphy.gif&ct=g

jmcouffin: https://media.giphy.com/media/Kzu05A116shtnT436d/giphy.gif

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Preflight Checks.pushbutton/config.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: UTF-8 -*-
+
+from os.path import dirname, abspath, join
+from pyrevit import script
+from pyrevit.forms import alert, pick_file, pick_folder
+from csv import reader
+
+script.get_output().close_others()
```
sanzoghenzo: Is this really needed? this doesn't open any output window after all...

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Preflight Checks.pushbutton/config.py
```diff
@@ -0,0 +1,33 @@
...
+
+config = script.get_config()
+
+current_folder = dirname(abspath(__file__))
+config.set_option(op_name='current_folder', value=current_folder)
+
+critical_warnings_filepath = pick_file(init_dir=current_folder,
+                                             file_ext="csv",
+                                             multi_file=False, 
+                                             title="Select the CSV file containing the critical warnings GUIDs")
```
sanzoghenzo: fix the formatting. This is the perfect example of why I personally like black/ruff formatting more: no need to move the arguments around if you change the variable or function name...

```suggestion
critical_warnings_filepath = pick_file(
    init_dir=current_folder,
    file_ext="csv",
    multi_file=False, 
    title="Select the CSV file containing the critical warnings GUIDs"
)
```

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Preflight Checks.pushbutton/config.py
```diff
@@ -0,0 +1,33 @@
...
+if critical_warnings_filepath is None:
+    alert("No CSV file selected. Exiting...", exitscript=True)
```
sanzoghenzo: Do we need to create an alert if the user consciously selected "cancel" in the pick file dialog?
```suggestion
    script.exit()
```

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Preflight Checks.pushbutton/config.py
```diff
@@ -0,0 +1,33 @@
...
+
+with open(critical_warnings_filepath, 'r') as f:
+    critical_warnings = []
+    for line in reader(f):
+        critical_warnings.append(line[0])
+    config.set_option(op_name='critical_warnings', value=critical_warnings)
+
+export_file_path = pick_folder(title="Where to save the data that will be collected?")
+if export_file_path is None:
+    alert("No file selected. Exiting...", exitscript=True)
```
sanzoghenzo:
```suggestion
    script.exit()
```

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Preflight Checks.pushbutton/config.py
```diff
@@ -0,0 +1,33 @@
...
+config.set_option(op_name='export_file_path', value=join(export_file_path+'\\projects_data.csv'))
```
sanzoghenzo: the `os.path.join` function exist exactly to avoid having to concatenate string and escape paths separators 😉 
```suggestion
config.set_option(op_name='export_file_path', value=join(export_file_path, 'projects_data.csv'))
```

sanzoghenzo (changes requested):
It's time to roast the config module 😅

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,396 @@
...
+        output.print_md('## ' + link_printed_name)
+        output.print_md('___')
+        if rvtlink is None:
+            continue
+        output.print_md(link_name)
+        element_count = elements_count(rvtlink)
+        purgeable_elements_count = get_purgeable_count(rvtlink)
+        all_warnings_count, _, warnings_guid = warnings(rvtlink)
+        critical_warnings_count = critical_warnings(warnings_guid,CRITICAL_WARNINGS)
+        rvtlinks_elements_items, rvtlinks_count, type_status, rvtlinks_docs = rvtlinks_elements(rvtlink)
```
sanzoghenzo: here you're rewriting the `rvtlinks_docs` list that is used in the for loop, is this the intended behavior?
I believe the quickest solution is to rename the function argument and line 241 to something else

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,396 @@
...
+        text_notes_types_count, text_notes_types_wf_count, text_bg_count = text_notes_types(rvtlink)
+        text_notes_count, text_notes_caps_count = text_notes_instances(rvtlink)
+        detail_groups_count, detail_groups_types_count = detail_groups(rvtlink)
+        model_group_count, model_group_type_count = groups(rvtlink)
+        reference_planes_count, unnamed_ref_planes_count = reference_planes(rvtlink)
+        detail_lines_count = detail_lines(rvtlink)
+        dim_types_count, dim_count, dim_overrides_count = count_dimension_types(rvtlink), count_dimensions(rvtlink), count_dimension_overrides(rvtlink)
+        revision_clouds_count = revisions_clouds(rvtlink)
+        links_data = ''
+        if rvtlinks_count != 0:
```
sanzoghenzo: same thing as the main function, just use 
```suggestion
        if rvtlinks_elements_items:
```

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,396 @@
...
+            links_data += card_builder(50, rvtlinks_count, ' Links') + \
+                        card_builder(0, rvtlinks_unpinned,
+                                     ' Links not pinned')
```
sanzoghenzo: links_data is empty at this point, so no need to concatenate...
```suggestion
            links_data = (
                card_builder(50, rvtlinks_count, ' Links') + 
                card_builder(0, rvtlinks_unpinned, ' Links not pinned')
            )
```

sanzoghenzo (changes requested):
Another quick go at the main script.

Since this is a new file, I would start to enforce a bit of formatting rules.
[black has good defaults](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#how-black-wraps-lines) and it is already installed in the pipenv, so you just need to 

```shell
pipenv run black extensions/pyRevitTools.extension/checks/audit_all_check.py
```

pyrevitlib/pyrevit/preflight/query.py
```diff
@@ -0,0 +1,1052 @@
...
+from pyrevit import DB, script, forms
+from pyrevit.revit.db.query import get_name
+from pyrevit.framework import get_type
+
+
+config = script.get_config()
+if config is None:
+    forms.alert("No config file found. Exiting...", exitscript=True)
+
+
```
sanzoghenzo: why is this here? config is not used anywhere...
```suggestion
```

pyrevitlib/pyrevit/preflight/query.py
```diff
@@ -0,0 +1,1052 @@
...
+        str: A comma-separated list of the names of the phases in a project.
+    """
+    if not hasattr(document, "Phases"):
+        return '-'
+    phases_collection = document.Phases
+    phase_names = []
+    for phase in phases_collection:
+        phase_names.append(phase.Name)
+    project_phases = (", ").join(phase_names)
+    return project_phases
```
sanzoghenzo: this could be a one-liner
```suggestion
    return ", ".join(phase.Name for phase in document.Phases)
```

pyrevitlib/pyrevit/preflight/query.py
```diff
@@ -0,0 +1,1052 @@
...
+    Args:
+        document (Document): A Revit document.
+
+    Returns:
+        str: A string with the names of all user worksets in a document.
+    """
+
+    if not hasattr(document, "IsWorkshared"):
+        return '-'
+    elif not document.IsWorkshared:
```
sanzoghenzo:
```suggestion
    if not document.IsWorkshared:
```

pyrevitlib/pyrevit/preflight/query.py
```diff
@@ -0,0 +1,1052 @@
...
+        return 'Not Workshared'
+    worksets_collection = (
+        DB.FilteredWorksetCollector(document)
+        .OfKind(DB.WorksetKind.UserWorkset)
+        .ToWorksets()
+        )
+    worksets_names = ", ".join(w.Name for w in worksets_collection)
+    return worksets_names
```
sanzoghenzo:
```suggestion
    return ", ".join(w.Name for w in worksets_collection)
```

pyrevitlib/pyrevit/preflight/query.py
```diff
@@ -0,0 +1,1052 @@
...
+    """
+    Returns a list of all rvt_links instances in a document
+
+    Args:
+        document (Document): A Revit document.
+
+    Returns:
+        list: A list of Revit link instances.
+    """
+    return (DB.FilteredElementCollector(document).OfCategory(DB.BuiltInCategory.OST_RvtLinks).WhereElementIsNotElementType())
```
sanzoghenzo:
```suggestion
    return DB.FilteredElementCollector(document).OfCategory(DB.BuiltInCategory.OST_RvtLinks).WhereElementIsNotElementType()
```

pyrevitlib/pyrevit/preflight/query.py
```diff
@@ -0,0 +1,1052 @@
...
+
+    Args:
+        revitlinks_elements (list): A list of Revit link elements.
+
+    Returns:
+        tuple: Two lists of Revit document names and Revit link instances names.
+    """
+    rvt_links_docs_name = [get_name(rvtlinks_element).split(' : ')[0].split('.rvt')[0] for rvtlinks_element in revitlinks_elements]
+    rvt_links_instances_name = [get_name(rvtlinks_element).split(' : ')[1] for rvtlinks_element in revitlinks_elements]
+    return rvt_links_docs_name, rvt_links_instances_name
```
sanzoghenzo: I'm not convinced on that one, you do the same operation on the elements of a list twice to get two different parts of a string... this in one of the cases in which a normal loop makes more sense

pyrevitlib/pyrevit/preflight/query.py
```diff
@@ -0,0 +1,1052 @@
...
+    param = DB.BuiltInParameter.STRUCTURAL_ANALYTICAL_MODEL
+    analytical_elements = (
+        DB.FilteredElementCollector(document)
+        .WhereElementIsNotElementType()
+        .ToElements()
+        )
+    count = 0
+    for element in analytical_elements:
+        if element.get_Parameter(param):
+            if element.get_Parameter(param).AsInteger() == 1:
```
sanzoghenzo:
```suggestion
        if element.get_Parameter(param) and element.get_Parameter(param).AsInteger() == 1:
```

pyrevitlib/pyrevit/preflight/query.py
```diff
@@ -0,0 +1,1052 @@
...
+    int: The number of views in the set that have "Copy" or "Copie" in their name.
+    """
+    copied_views_count = 0
+    for view in views_set:
+        view_name = get_name(view)
+        try:
+            if "Copy" in view_name or "Copie" in view_name:
+                copied_views_count += 1
+        except AttributeError:
+            pass
```
sanzoghenzo: how can this block raise an AttributeError?

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Preflight Checks.pushbutton/config.py
jmcouffin: Old habit, I run tools so many times that I get dozens of output windows. This allows me to start with a clean slate.

pyrevitlib/pyrevit/preflight/query.py
```diff
@@ -0,0 +1,1052 @@
...
+        title (str): The title to be displayed on the top left corner of the frame.
+        cards (str): Multiple strings representing HTML card elements.
+    Returns:
+        str: A string containing the HTML representation of the div frame.
+    """
+    frame_start = '<div style="display: inline-block; border: 1px solid #ccc; border-radius: 10px; padding: 5px; margin: 10px 5px 10px 5px; position: relative;">'
+    title_html = '<div style="position: absolute; top: -10px; left: 10px; background: white; padding: 0 5px; font-weight: bold;">{}</div>'.format(title)
+    cards_html = ''.join(cards)
+    frame_end = '</div>'
+    return frame_start + title_html + cards_html + frame_end
```
sanzoghenzo: these are not queries, should be moved to another module.
Maybe add it to the pyrevit.output module for everyone to enjoy?

sanzoghenzo (changes requested):
Just a few comments on the query, but this is because it's late...

In general, I have a few points to make:
- why return a list of item and it's count instead of using len(items) where it is needed?
- the `ToElement()` should never be None, at most it is an empty list, so all the `is None` checks are useless. Or am I missing something?
- use pyrevit.revit.db.query module more :wink: like `get_elements_by_class`... and if there are missing functionality, add it there

jmcouffin:
> ```shell
> pipenv run black extensions/pyRevitTools.extension/checks/audit_all_check.py
> ```

And this is how, once again, you made my day.
Why wasn't using this before... ?
Thanks
@sanzoghenzo

jmcouffin:
> > ```shell
> > pipenv run black extensions/pyRevitTools.extension/checks/audit_all_check.py
> > ```
> 
> And this is how, once again, you made my day. Why wasn't using this before... ? Thanks @sanzoghenzo

I get it why you want to make it part of the CI actions!

sanzoghenzo:
@jmcouffin  I'm having an hard time following what you're doing... You already merged an edit on the pyrevit query module that I would have rejected 😅 

Sorry for being too pedantic, but the `get_phases_names` as it is is too specific to this case; It should return a list of names, not the names already joined in a CSV for it to be useful to everybody.

Please do your edits without merging anything else, then when you're done use the "Ready for review" button to notify me and wait for my final approval to merge (it may be that after solving the review other things come up, so another review cycle is in order, and so on...)

jmcouffin:
> You already merged an edit on the pyrevit query module that I would have rejected 😅

That is because, when I started trying to reuse the query module, I noticed the missing docstring then I thought, while at it... Seeing the length of the module, I realised it would have been better done in a different PR _ even though I already started to add my functions... I know... not ideal. 
Russian doll, I keep in seeing things I should fix, then I get knee deep in it!


> Sorry for being too pedantic, but the `get_phases_names` as it is is too specific to this case; It should return a list of names, not the names already joined in a CSV for it to be useful to everybody.

Not at all. I take all the critics. Not offended. I am good willing, and unperfect ;p
will fix later and split it in bits that are more useful



> Please do your edits without merging anything else, then when you're done use the "Ready for review" button to notify me and wait for my final approval to merge (it may be that after solving the review other things come up, so another review cycle is in order, and so on...)

🫡

sanzoghenzo:
> Russian doll, I keep in seeing things I should fix, then I get knee deep in it!

I feel you, I can't count how many times I reformatted the code I was fixing, only to revert it back to avoid too much pollution in the PR... 😉

jmcouffin:
> Just a few comments on the query, but this is because it's late...
> 
> In general, I have a few points to make:
> 
> * [x]  why return a list of item and it's count instead of using len(items) where it is needed?
> * [x]  the `ToElement()` should never be None, at most it is an empty list, so all the `is None` checks are useless. Or am I missing something?
> * [x]  use pyrevit.revit.db.query module more 😉 like `get_elements_by_class`... and if there are missing functionality, add it there

because it is late and I spent the whole evening finishing that up, Thanks. I am learning to be a better baby programmer!

jmcouffin:
I will finish up by fixing some of the numbers that are wrong by doing a last QC by the end of the week

jmcouffin:
@devloai review PR

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Preflight Checks.pushbutton/config.py
```diff
@@ -0,0 +1,46 @@
...
+critical_warnings_filepath = pick_file(
+    init_dir=current_folder,
+    file_ext="csv",
+    multi_file=False,
+    title="Select the CSV file containing the critical warnings GUIDs",
+)
+if critical_warnings_filepath is None:
+    script.exit()
+
+with open(critical_warnings_filepath, "r") as f:
```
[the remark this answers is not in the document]
jmcouffin: IPY does not not support encoding specification with open function

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,745 @@
...
+    def export_to_csv(self, export_file_path=EXPORT_FILE_PATH, headers=None):
+        """
+        Export data to a CSV file. If the file does not exist, 
+        it will be created. If the file exists, 
+        the data will be appended based on the date and document name.
+        """
+        if headers is None:
+            headers = self.COLUMNS
+        if not isfile(export_file_path):
+            with open(export_file_path, mode="wb") as csv_file:
```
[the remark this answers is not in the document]
jmcouffin: not supported in IPY

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,745 @@
...
+            [project_info],
+            columns=FILE_INFO_HEADERS,
+        )
+
+        # Linked files infos
+        links_cards = ""
+        if current_doc_report.rvtlinks_elements_items:
+            link_data = []
+            rvtlinks_instances = get_linked_model_instances(doc).ToElements()
+            rvtlinks_documents = (link.GetLinkDocument() for link in rvtlinks_instances)
```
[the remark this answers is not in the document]
jmcouffin: solved everything except copied view

jmcouffin:
Hi @sanzoghenzo 
That puppy is ready for your review. I put a lot of love into this one. Hoping this won't be too painful to look at 😄

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Preflight Checks.pushbutton/config.py
sanzoghenzo: Ok, but... this doesn't open any new window, to me it shouldn't care about closing the others

pyrevitlib/pyrevit/output/cards.py
```diff
@@ -0,0 +1,90 @@
...
+    color = "#d0d3d4"
+    if value != 0:
+        if 0.5 <= ratio <= 1:
+            color = "#FFDD94"  # orange
+        elif ratio > 1:
+            color = "#FA897B"  # red
+        elif ratio < 0.5:
+            color = "#D0E6A5"  # green
+        else:
+            pass
```
sanzoghenzo:
```suggestion
        if ratio < 0.5:
            color = "#D0E6A5"  # green
        elif ratio <= 1:
            color = "#FFDD94"  # orange
        else:
            color = "#FA897B"  # red
```

pyrevitlib/pyrevit/output/cards.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+
+    alt = "{} {} (limit = {})".format(str(value), str(description), str(limit))
+    card_end = (
+        "<b>"
+        + str(value)
+        + '</b><br /><a style="font-size: 0.70rem">'
+        + description
+        + "</a></div>"
+    )
```
sanzoghenzo:
```suggestion
    card_end = '<b>{}</b><br /><a style="font-size: 0.70rem">{}</a></div>'.format(
        value, description
    )
```

pyrevitlib/pyrevit/output/cards.py
```diff
@@ -0,0 +1,90 @@
...
+    card = card_start_style(limit, value, alt) + card_end
+    return card
```
sanzoghenzo:
```suggestion
    return card_start_style(limit, value, alt) + card_end
```

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -730,42 +729,64 @@ def get_worksets_names(doc=None):
...
 
-    Returns the number of warnings in the document
+    Args:
+        doc (Document): A Revit document.
 
+    Returns:
+        list: All the warnings in the document.
+    """
+    doc = doc or DOCS.doc
+    return doc.GetWarnings() if doc.GetWarnings() else []
```
sanzoghenzo: the correct/pythonic way is
```suggestion
    return doc.GetWarnings() or []
```

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -730,42 +729,64 @@ def get_worksets_names(doc=None):
...
+        int: Number of warnings in the document.
     """
     doc = doc or DOCS.doc
-    all_warnings = doc.GetWarnings()
-    if not all_warnings:
-        return 0, [], []
-    warnings_guid = [warning.GetFailureDefinitionId().Guid for warning in all_warnings]
-    return len(all_warnings), all_warnings, warnings_guid
+    all_warnings = get_warnings(doc)
+    return len(all_warnings) if all_warnings else 0
```
sanzoghenzo: get_warnings always return a list, so the check is useless. and since this all comes down to `len(get_warnings(doc))`, is a separate function really needed? you only use it once after all...

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -730,42 +729,64 @@ def get_worksets_names(doc=None):
...
+        doc (Document): A Revit document.
 
-    Parameters:
-    warnings_guid (list): A list of warning GUIDs.
+    Returns:
+        list: A list of GUIDs for all the warnings in the document.
+    """
+    doc = doc or DOCS.doc
+    all_warnings = get_warnings(doc)
+    return [warning.GetFailureDefinitionId().Guid for warning in all_warnings] if all_warnings else []
```
sanzoghenzo:
```suggestion
    return [warning.GetFailureDefinitionId().Guid for warning in get_warnings(doc)]
```

then again, you only use this once in `get_critical_warning_count` below, isn't it better to pass the warnings to it and extract the guids inside it?

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -1038,6 +1058,35 @@ def get_sheets(include_placeholders=True, include_noappear=True, doc=None):
     return sheets
 
 
+def get_document_clean_name(doc=None):
```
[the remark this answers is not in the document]
sanzoghenzo: The comment was good, even though python 2.7 doesn't have pathlib, pyrevitlib ships with its own version that can be used. I see you used the `join` function but you don't join nothing, why?

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -1123,6 +1167,77 @@ def get_linked_model_doc(linked_model):
...
+    Returns:
+        list: A list of statuses for each linked Revit model type.
+    """
+    doc = doc or DOCS.doc
+    rvtlinks_instances = get_linked_model_instances(doc)
+    rvtlinks_types = get_linked_model_types(doc, rvtlinks_instances)
+    return [rvtlinktype.GetLinkedFileStatus() for rvtlinktype in rvtlinks_types]
+
+
+def get_rvt_link_doc_name(rvtlink_instance=None):
```
sanzoghenzo: get_name needs exactly one argument, so should also this function

```suggestion
def get_rvt_link_doc_name(rvtlink_instance):
```

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -1123,6 +1167,77 @@ def get_linked_model_doc(linked_model):
...
+    Args:
+        rvtlink_instance: The Revit link instance from which to extract the document name.
+    
+    Returns:
+        str: The name of the Revit link document, without the file extension and any directory paths.
+    """
+    return get_name(rvtlink_instance).split(" \ ")[0].split(".rvt")[0]
+
+
+def get_rvt_link_instance_name(rvtlink_instance=None):
```
sanzoghenzo, line 1184:
```suggestion
def get_rvt_link_instance_name(rvtlink_instance):
```

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -3076,3 +3209,38 @@ def get_geometry(element, include_invisible=False, compute_references=False):
...
+    """
+    array_list = DB.FilteredElementCollector(doc or DOCS.doc).OfCategory(
+        DB.BuiltInCategory.OST_IOSArrays
+    )
+    arrays_groups = []
+    for ar in array_list:
+        all_arrays_members = list(ar.GetOriginalMemberIds())
+        all_arrays_members.extend(list(ar.GetCopiedMemberIds()))
+        arrays_groups.append(all_arrays_members)
+    return [item for sublist in arrays_groups for item in sublist]
```
sanzoghenzo: This wastes RAM, you can just extend the `arrays_groups` right away
```suggestion
        arrays_groups.extend(ar.GetOriginalMemberIds())
        arrays_groups.extend(ar.GetCopiedMemberIds()))
    return arrays_groups
```

pyrevitlib/pyrevit/revit/db/count.py
```diff
@@ -0,0 +1,762 @@
...
+    Returns the number of unbounded rooms in the document.
+
+    Args:
+        document (Document): A Revit document.
+
+    Returns:
+        int: The number of unbounded rooms in the document.
+    """
+    rooms = get_elements_by_categories([DB.BuiltInCategory.OST_Rooms], doc=document)
+    return sum(1 for room in rooms if room.Area == 0)
```
sanzoghenzo: With this I believe you'll be mad at me 😅 
This also applies to most of the other functions in this module: you keep asking for the elements (rooms in this case) only to count them on a different condition. Change these function to accept the list of elements they need an move the elements retrieval in the caller

pyrevitlib/pyrevit/revit/db/count.py
```diff
@@ -0,0 +1,762 @@
...
+    Counts the total number of schedules in the given document.
+
+    Args:
+        document (DB.Document): The Revit document to query for schedules.
+
+    Returns:
+        int: The total number of schedules in the document.
+    """
+    schedules_elements = get_all_schedules(document)
+    return len(schedules_elements)
```
sanzoghenzo: do you really need a function for a `len`? 😉

pyrevitlib/pyrevit/revit/db/count.py
```diff
@@ -0,0 +1,762 @@
...
+    Args:
+    - document (DB.Document): The Revit document to search for parameter filters.
+
+    Returns:
+    - int: The count of all parameter filters in the document.
+    """
+    filters_collection = get_elements_by_class(DB.ParameterFilterElement, doc=document)
+    if not filters_collection:
+        return 0
+    return len(filters_collection)
```
sanzoghenzo:
```suggestion
    len(get_elements_by_class(DB.ParameterFilterElement, doc=document) or [])
```

but again, no need for a function to do this

pyrevitlib/pyrevit/revit/db/count.py
```diff
@@ -0,0 +1,762 @@
...
+    Args:
+        document (DB.Document): The Revit document to search for DWG files.
+
+    Returns:
+        int: The total number of DWG files in the document.
+    """
+    dwg_collector = get_elements_by_class(DB.ImportInstance, doc=document)
+    if not dwg_collector:
+        return 0
+    return len(dwg_collector)
```
sanzoghenzo:
```suggestion
    return len(get_elements_by_class(DB.ImportInstance, doc=document) or [])
```

same as above

pyrevitlib/pyrevit/revit/db/count.py
```diff
@@ -0,0 +1,762 @@
...
+    dimensions = get_elements_by_categories(
+        [DB.BuiltInCategory.OST_Dimensions], doc=document
+    )
+    if not dimensions:
+        return 0
+    dimension_count = 0
+    for d in dimensions:
+        if d.OwnerViewId and d.ViewSpecific and d.View:
+            dimension_count += 1
+    return dimension_count
```
sanzoghenzo:
```suggestion
    return sum(1 for d in dimensions if d.OwnerViewId and d.ViewSpecific and d.View)
```

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,983 @@
...
+    count_elements,
+    count_detail_lines,
+    count_dimensions,
+    count_dimension_types,
+    count_dimension_overrides,
+    count_revision_clouds,
+    count_purgeable_elements,
+    count_generic_models_types,
+)
+
```
sanzoghenzo: provided the other comments, instead of importing every single function import, and maybe alias, the module itself
Obviously you need to update all the calls by prepending the module name

```suggestion
from pyrevit.revit.db import query # as q?
from pyrevit.preflight import PreflightTestCase
from pyrevit.revit.db import count # as cnt?
```

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,983 @@
...
+        7. Generates and prints an HTML report.
+        8. Exports the report data to a CSV file.
+        9. Generates a detailed report for RVTLinks if present.
+
+    Exceptions:
+        Logs errors and prints the stack trace if any exception occurs during the process.
+    """
+    try:
+        if doc.IsFamilyDocument:
+            alert("This tool is for project files only. Exiting...", exitscript=True)
```
sanzoghenzo: I would let the TestCase handle the teardown instead of exiting early

```suggestion
            alert("This tool is for project files only. Exiting...")
            return
```

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,983 @@
...
+                        data.document_info.clean_name, warnings_count
+                    ),
+                    is_favourite=True,
+                    is_new=True,
+                )
+            except Exception as e:
+                logger.error(
+                    "Failed to show balloon notification, Exception: {}".format(e),
+                    exc_info=True,
+                )
```
sanzoghenzo: In which cases the balloon doesn't work? can we be more specific about the exception?

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,983 @@
...
+            "N/A",
+            "N/A",
+            "N/A",
+        ]
+        output.print_md("# Main File Infos")
+        output.print_table([project_info], columns=FILE_INFO_HEADERS)
+
+        # Linked files infos
+        links_cards = ""
+        if data.links_info.rvtlinks_count > 0:
```
sanzoghenzo: help me understand why you use the count and then get the elements instead of just retrieving the elements and do a `len` on them (see the other comments)...

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,983 @@
...
+            link_data = []
+            rvtlinks_instances = get_linked_model_instances(doc).ToElements()
+            rvtlinks_documents = (link.GetLinkDocument() for link in rvtlinks_instances)
+            links_documents_data = [
+                ReportData(link_doc) for link_doc in rvtlinks_documents
+            ]
+            for rvt_link_instance, link_document_data in zip(
+                rvtlinks_instances, links_documents_data
+            ):
```
sanzoghenzo: If I get it right, this can be made into a single loop
```suggestion
            link_data = []
            for rvt_link_instance in get_linked_model_instances(doc).ToElements():
                link_doc = rvt_link_instance.GetLinkDocument()
                link_document_data = ReportData(link_doc)

```

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,983 @@
...
+
+class DocumentInfo:
+    """Handles document-specific information."""
+
+    def __init__(self, document):
+        self.clean_name = get_document_clean_name(document) if document else ""
+        self.revit_version_build = HOST_APP.build
+
+
+class ProjectInfoData:
```
sanzoghenzo, line 70: I'm worried about this repeated checks, I believe the existence of the document should be delegated to the caller (I can't imagine a scenario where all these info are requested on a non existing document)

sanzoghenzo (commented):
I may have gone too hard on some things, feel free to pick whatever advice you want 😉

jmcouffin:
> I may have gone too hard on some things, feel free to pick whatever advice you want 😉

Never mad. This is all constructive criticism.
Will look at it later
Thanks

extensions/pyRevitTools.extension/checks/audit_all_check.py
jmcouffin: I wouldn't know ;p

extensions/pyRevitTools.extension/checks/audit_all_check.py
jmcouffin: makes sense

extensions/pyRevitTools.extension/checks/audit_all_check.py
jmcouffin: ok, I hate you for this one ;p

jmcouffin: Honestly, I felt it was cleaner, but this is jsut the noob speaking

pyrevitlib/pyrevit/revit/db/count.py
jmcouffin: 😄definitly not mad

jmcouffin:
I went through the whole thing! Damned... a beast of a refactoring. I learned a lot. Thanks @sanzoghenzo 

I am ready to merge except for that one SOB groups counting... for the life of me, I spend way too much time on this. 
in a very simple homemade sample, the count is right, in the revit arch sample it is off...

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -1123,6 +1123,77 @@ def get_linked_model_doc(linked_model):
                 return open_doc
 
 
+def get_linked_model_types(doc=None, rvt_links_instances=None):
```
sanzoghenzo:
```suggestion
def get_linked_model_types(doc, rvt_links_instances):
```

pyrevitlib/pyrevit/revit/db/query.py
```diff
@@ -730,42 +729,20 @@ def get_worksets_names(doc=None):
     return ", ".join(w.Name for w in worksets_collection)
 
 
-def get_warnings_info(doc=None):
+def get_critical_warnings_count(warnings=None, critical_warnings_template=None):
```
sanzoghenzo:
```suggestion
def get_critical_warnings_count(warnings, critical_warnings_template):
```

pyrevitlib/pyrevit/revit/db/count.py
```diff
@@ -0,0 +1,455 @@
...
+    rule = DB.FilterIntegerRule(provider, evaluator, 1)
+    elem_param_filter = DB.ElementParameterFilter(rule)
+    return (
+        DB.FilteredElementCollector(document)
+        .WherePasses(elem_param_filter)
+        .GetElementCount()
+    )
+
+
+def count_unplaced_schedules(schedules=None):
```
sanzoghenzo:
```suggestion
def count_unplaced_schedules(schedules):
```

pyrevitlib/pyrevit/revit/db/count.py
```diff
@@ -0,0 +1,455 @@
...
+    Args:
+        sheets_set (list): A list of sheets.
+        views_count (int): The number of views in the document.
+
+    Returns:
+        int: The number of views not on a sheet.
+    """
+    views_on_sheet = []
+    if sheets_set is None or views_count is None:
+        return views_count
```
sanzoghenzo, line 67: This may return None, not an integer... it is expected?

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,925 @@
...
+    Methods:
+        __init__(self, document=None):
+            Initializes the ReportData instance with the provided Revit document.
+        COLUMNS(self):
+            Returns a list of column names based on the keys in _KEY_MAPPING.
+        to_list(self):
+        export_to_csv(self, export_file_path=EXPORT_FILE_PATH, headers=None):
+    """
+
+    _KEY_MAPPING = OrderedDict(
```
sanzoghenzo, line 344: this could be a module-level constant, but since you use the class only once it is not that bad to leave it here.

extensions/pyRevitTools.extension/checks/audit_all_check.py
```diff
@@ -0,0 +1,925 @@
...
+            ),
+            ("detail_lines_count", "family_info.detail_lines_count"),
+            ("dim_types_count", "dimensions_info.dim_types_count"),
+            ("dim_count", "dimensions_info.dim_count"),
+            ("dim_overrides_count", "dimensions_info.dim_overrides_count"),
+            ("revision_clouds_count", "revisions_info.revision_clouds_count"),
+        ]
+    )
+
+    def __init__(self, document=None):
```
sanzoghenzo:
```suggestion
    def __init__(self, document):

```

sanzoghenzo (changes requested):
Last one: since there are no None checks in place, avoid having the function arguments default to it

sanzoghenzo (dismissed):
Looks good to me!

jmcouffin:
The puppy is out in the wild.
Thank you so much @sanzoghenzo for the comments!

Issue #2512, opened 2025-01-12
[Bug]: pyRevit_config.ini is being used by another process


archeduardo5:
### ✈ Pre-Flight checks

- [X] I **don't have SentinelOne** antivirus installed (see above for the solution)
- [X] I have **searched in the issues** (open and closed) but couldn't find a similar issue
- [X] I have **searched in the [pyRevit Forum](https://discourse.pyrevitlabs.io)** for similar issues
- [X] I already **followed the [installation troubleshooting guide](https://pyrevitlabs.notion.site/Installation-issues-f6495adb56254e9e8021f8a68e3b2ab7) thoroughly**
- [X] I am using the **latest pyRevit Version**

### 🐞 Describe the bug

PyRevit is unable to load due to a file access conflict with the configuration file. When launching Revit, pyRevit throws an IOError indicating that the pyRevit_config.ini file is being used by another process and cannot be accessed.

This happens on every Revit startup attempt and prevents pyRevit from loading properly. The error occurs specifically with the configuration file located at:
C:\Users\[username]\AppData\Roaming\pyRevit\pyRevit_config.ini

The error prevents PyRevit from initializing and loading any of its tools or functions. This appears to be a file locking issue where the configuration file is either:
1. Locked by another process
2. Has incorrect permissions
3. Is being accessed simultaneously by multiple processes

### ⌨ Error/Debug Message

```shell
IronPython Traceback:
Traceback (most recent call last):
  File "C:\Program Files\pyRevit-Master\bin\engines\pyRevitLoader.py", line 33, in <module>
  File "C:\Program Files\pyRevit-Master\pyrevitlib\pyrevit\loader\sessionmgr.py", line 267, in load_session
  File "C:\Program Files\pyRevit-Master\pyrevitlib\pyrevit\loader\sessioninfo.py", line 49, in setup_runtime_vars
  File "C:\Program Files\pyRevit-Master\pyrevitlib\pyrevit\userconfig.py", line 689, in get_current_attachment
IOError: [Errno 32] The process cannot access the file 'C:\Users\arch1\AppData\Roaming\pyRevit\pyRevit_config.ini' because it is being used by another process.

Script Executor Traceback:
System.IO.IOException: The process cannot access the file 'C:\Users\arch1\AppData\Roaming\pyRevit\pyRevit_config.ini' because it is being used by another process.
   at Microsoft.Scripting.Runtime.LightExceptions.ThrowException(LightException lightEx)
[...]

PR #2516, opened 2025-01-13
fix(config): allow concurrent access to pyrevit_config.ini

### LINKED ISSUE #2463: [Bug]: manage access to pyrevit-config.ini at startup
### ✈ Pre-Flight checks

### 🐞 Describe the bug

- When launching two or more Revit sessions at the same time, pyrevit-config.ini gets accessed at the same time and throws error
- pyRevit cannot be launched for these sessions except for the one that caught the config file first
https://discourse.pyrevitlabs.io/t/failed-to-save-config-in-use-by-another-program/8187/3

image

### ⌨ Error/Debug Message

```shell
'used by another process'
```

### ♻️ To Reproduce

Launch multiple sessions of Revit at once (within 10/30sec)

### ⏲️ Expected behavior

- be able to launch as many revit sessions as possible at the same time
- handling access to pyrevit-config.ini properly so that it does not get locked

```shell
NA
```

### Additional context

old issue, never fixed

sanzoghenzo:
fixes #2463 #2512

It seems to work on my machine, I ran 3 instances of revit 2023 one after another, and then 2 instances of revit 2025.

### DIFF
--- a/dev/pyRevitLabs/pyRevitLabs.PyRevit/PyRevitConfig.cs
+++ b/dev/pyRevitLabs/pyRevitLabs.PyRevit/PyRevitConfig.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
 
 using pyRevitLabs.Common;
 using pyRevitLabs.Common.Extensions;
@@ -31,7 +32,9 @@ namespace pyRevitLabs.PyRevit {
                 cfgOps.Encoding = CommonUtils.GetUTF8NoBOMEncoding();
                 _config = new IniFile(cfgOps);
 
-                _config.Load(cfgFilePath);
+                using (var cfgStream = File.Open(cfgFilePath, FileMode.Open, FileAccess.Read, FileShare.Read)) {
+                    _config.Load(cfgStream);
+                }
                 _adminMode = adminMode;
             }
             else


jmcouffin (approved):

jmcouffin:
I launched 3x2024 within seconds, no issues
3x2023, 1 out of 3 greeted me with 
image
This is a great improvement

jmcouffin:
exception raised here https://github.com/sanzoghenzo/pyRevit/blob/cc11f7680b8bea25bee8360c318445cfbbd6026e/dev/pyRevitLabs/pyRevitLabs.PyRevit/PyRevitConfig.cs#L53

jmcouffin:
Me and my friend ChatG got this working using an ugly retry mechanism:
```C#
        using System.IO;
        
        //...


        // save config file to standard location
        public void SaveConfigFile()
        {
            if (_adminMode)
            {
                logger.Debug("Config is in admin mode. Skipping save");
                return;
            }

            logger.Debug("Saving config file \"{0}\"", PyRevitConsts.ConfigFilePath);

            int retryCount = 5;            // Number of retry attempts
            int delayMilliseconds = 100;   // Delay between retries

            while (retryCount > 0)
            {
                try
                {
                    _config.Save(PyRevitConsts.ConfigFilePath);
                    return; // Exit method if successful
                }
                catch (IOException ex) // Catch only IO exceptions related to file locking
                {
                    retryCount--;
                    if (retryCount == 0)
                    {
                        throw new PyRevitException(string.Format("Failed to save config to \"{0}\" after multiple attempts. | {1}",
                                                                 PyRevitConsts.ConfigFilePath, ex.Message));
                    }

                    logger.Debug("Retrying to save config file \"{0}\"... Remaining retries: {1}", 
                                 PyRevitConsts.ConfigFilePath, retryCount);

                    System.Threading.Thread.Sleep(delayMilliseconds); // Wait before retrying
                }
                catch (Exception ex)
                {
                    throw new PyRevitException(string.Format("Failed to save config to \"{0}\". | {1}",
                                                             PyRevitConsts.ConfigFilePath, ex.Message));
                }
            }
        }

```
Thoughts @sanzoghenzo, except that it is a crappy way of doing it ;p

sanzoghenzo:
Why does the "get current attachments" need to save the file? This clones and attachments thing starts to driving me mad 🤯 

From a quick look at it, I believe this is the culprit: 
https://github.com/pyrevitlabs/pyRevit/blob/6c94474cb68cff9469c02c337a179aab3a3d749b/dev/pyRevitLabs/pyRevitLabs.PyRevit/PyRevitClones.cs#L124

We need to write back only if the list of attachments have changed.

But in any case yes, we need to add some checks in the save method to handle this kind of concurrency.
But anyway, who's the crazy person that starts multiple revit instances at the same time?? 🤣

jmcouffin:
> But anyway, who's the crazy person that starts multiple revit instances at the same time?? 🤣

Trust me, plenty of BIM Managers do when they review files in Revit!



> From a quick look at it, I believe this is the culprit:
> 
> https://github.com/pyrevitlabs/pyRevit/blob/6c94474cb68cff9469c02c337a179aab3a3d749b/dev/pyRevitLabs/pyRevitLabs.PyRevit/PyRevitClones.cs#L124
> 
> We need to write back only if the list of attachments have changed.

Want to do it @sanzoghenzo ?

sanzoghenzo:
I'll try it, not sure if I can do this today;

For the retry solution you proposed, this mitigates the problem but doesn't solve the root issue: what to do when multiple instances try to write on the same file? It's the whole git/cloud storage problem again and again...
Let's see if we can reduce the problem as much as possible.

jmcouffin:
> but doesn't solve the root issue:

Agree 💯

PR #2517, opened 2025-01-13
docs: make the text in the logo follow the site theme

sanzoghenzo:
this one was bugging me for a long time...
I use dark mode almost everywhere, and the logo in the README/main docs page rendered awfully.

I added the `fill="currentColor"` attribute to the SVG texts, but in order to make it work I had to directly insert the image inside the markdown file, so now there are 63 lines of undreadable garbage at the beginning of the .md file 😅 

Up to you to decide if move this forward or not.

to check it locally, you can give the usual `pipenv mkdocs serve` command.


jmcouffin:
Question, 
when building locally, I do have to `pip install mkdocs` before running `pipenv mkdocs serve`
shouldn't it be in the pipfile/requirements already?

jmcouffin:
ugly solution but efficient!
💯 appoved 😛

PR #2523, opened 2025-01-15
query module docstring review

jmcouffin:
Used GH copilot to add missing docstrings and black to format everything @sanzoghenzo

No proof reading in the process, pure confidence in copilot 😓 but de reference API will be so much better/

The query module would need to be broken down in smaller bits but this is too much for now


sanzoghenzo:
@jmcouffin  if you don't want anyone to review things, then just merge your branch to develop without doing a PR since you can😉 

The docstring formatting is all wrong, there should be a blank line between sections (before `Args:`, `Returns:` and so on)

jmcouffin:
> The docstring formatting is all wrong, there should be a blank line between sections (before `Args:`, `Returns:` and so on)

you mean my copilot friend is dumb like me? ;p 
I will fix that

jmcouffin:
done

PR #2535, opened 2025-01-21
fix: incorrect handling of PYTHONPATH in SetupSearchPaths()

reformstudios:
# Fix incorrect handling of PYTHONPATH in SetupSearchPaths()

## Description

The current implementation of importing PYTHONPATH in to sys.path in the python environment treats PYTHONPATH as a single value. This isn't always the case. It also doesn't follow the official specification for python which describes the value of PYTHONPATH as a list of paths. 

---

## Checklist

  ```bash
  ```
- [ ] Changes are tested and verified to work as expected.

---

## Related Issues

- Resolves #[issue number]

---

## Additional Notes

---

### DIFF
--- a/dev/pyRevitLabs.PyRevit.Runtime/CPythonEngine.cs
+++ b/dev/pyRevitLabs.PyRevit.Runtime/CPythonEngine.cs
@@ -205,7 +205,10 @@ namespace PyRevitLabs.PyRevit.Runtime {
                 var paths = pythonPath.Split(Path.PathSeparator);
                 foreach (var path in paths)
                 {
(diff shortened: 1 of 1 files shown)


sanzoghenzo:
Looks good to me, and the suggested review is a nice addition.
We could also go a step further and avoid adding duplicated paths on the sysPaths list, but this is not a problem for python and could be done in a separate list.

I don't have any use case to test this, as I wrote in the forum I never touch the pythonpath, since it will mess with my (many) virtual environments.

sanzoghenzo (commented):
See my previous comment

jmcouffin:
Thanks for the input!

PR #2539, opened 2025-01-24
loading pyRevitLabs.json in PyrevitCLI manually if assembly is not found

thumDer:
# Name of your PR

## Description

I couldn't make the CLI work from the develop branch, it always resulted in
```powershell
❯ .\bin\pyrevit.exe env
==> Registered Clones (full git repos)
Error: Could not load file or assembly 'Newtonsoft.Json, Version=11.0.0.0, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.
Run with "--debug" option to see debug messages
```

The only way I've found to solve this issue is to explicitly force assembly resolution.

---

## Checklist

  ```bash
  ```

---


jmcouffin (approved):

thumDer:
Just for the record: I had a similar issue with another addin and it turned out one of the other dependencies had a dependency on a specific version of a specific assembly that had trouble loading, so I might have a look on all the other dependencies, if this is the case here.

PR #2548, opened 2025-01-31
tool: Update cad_audit_check.py

nasmovk:
# Update cad_audit_check.py

## Description
Fixed bug in WIP 5.0.0.25025
This follows my testing of the first implementation of my cad_audit_check into pyRevit, based on my original Hackathon submittal

 4. Removed from pyrevit.forms import alert (not needed)
26. Added class ButtonClass(Window), to manage my cancel button 

98. Added a condition to the get_load_stat(cad, is_link) function 
    if status == "Loaded": 
        return "Loaded"

---
Tested in Revit 2024 & 2025


jmcouffin:
@nasmovk can you check devlo's comment. After that I will merge

jmcouffin:
I am about to push the release button and would like to include this one in the release.

sanzoghenzo:
I would to the opposite as it has been done here, there's shouldn't be the need to create a class with a static method... just remove `ButtonClass.` form this line

https://github.com/pyrevitlabs/pyRevit/blob/71d3ac09f3bb35060ce7b37a2cee556689b58d40/extensions/pyRevitTools.extension/checks/cad_audit_check.py#L78

and remove ButtonClass altogether.

nasmovk:
I will check and update, if I can understand Andrea's remark.

sanzoghenzo:
> I will check and update, if I can understand Andrea's remark.

Sorry if I wasn't clear enough, I assumed that, since you put a class and a static method in the code, you knew the teminology :sweat_smile: 

1. remove rows 26 through 34
2. line 78 should become `Button("CANCEL", on_click=cancel_clicked),`

I thought I already reviewed that, but it may be lost in the merge frenzy of the model checker competition.

The StackOverflow answer you're referring to in the comments is overcomplicating a simpler solution: the OP had to just remove the parenthesis to make things work 
```
Button('Proceed', on_click=proceed_pressed)
```
instead of
```
Button('Proceed', on_click=proceed_pressed())
```
The latter calls the function before passing it to the button, while the former passes the function itself to the button, so that it is called when the button is pressed.

nasmovk:
I'm away skiing for the weekend, but I promise to look at this as soon as I'm back. Looking forward to the v5 launch.

PR #2557, opened 2025-02-03
Update cad_audit_check.py

nasmovk:
Corrected the cancel button issue.
Removed the unnecessary class

# Update cad_audit_check.py

## Description
The ButtonClass was not needed.

### DIFF
--- a/extensions/pyRevitTools.extension/checks/cad_audit_check.py
+++ b/extensions/pyRevitTools.extension/checks/cad_audit_check.py
@@ -66,7 +66,7 @@ def get_user_input():
         RadioButton("model", "in this project", True, GroupName="grp"), # GroupName implemented in class through kwargs
         RadioButton("active_view", "in only the active view", False, GroupName="grp"),
         Separator(),
-        Button("CANCEL", on_click=ButtonClass.cancel_clicked),
+        Button("CANCEL", on_click=cancel_clicked),
         Button("OK"),
     ]
 


jmcouffin:
I just merged both your PR @nasmovk 
Can you check the code on the develop branch looks fine?

nasmovk:
Just installedpyRevit 5.0.0.25034+1241-wip
Clicking the Preflight Checks button gave this error : 

Preflight Checks
No configuration set, run the Preflight
Checks configuration mode using
SHIFT+Click.

I don't think that is related my cad_audit_check

jmcouffin:
New feature ;p @nasmovk 
if you set it up and shift click then use the Full Model Audit you will get a cool preview

nasmovk:
OK. I understand now. But I think the UI ergonomics are not quite right.
This config message (choosing the location for projects_data.csv) should only appear if I want to run the Audit All check, 
The other checks don't use it I presume (?)
But it is a very nice feature in itself.

jmcouffin:
> OK. I understand now. But I think the UI ergonomics are not quite right. This config message (choosing the location for projects_data.csv) should only appear if I want to run the Audit All check, The other checks don't use it I presume (?) But it is a very nice feature in itself.

agreed, and I plan on adding the feature to other checks eventually, but this is what it is for now. Need to release pyRevit5 as is.

nasmovk:
Sure. It's not a criticism, of course.
My cad_audit_check works with the latest change.

But I have just spotted a small incomplete validation. It does not check if the current project has 0 CAD ImportInstance.
I can easily fix this, but I don't want to delay the release.

jmcouffin:
If you do it now now, I will include it.
@nasmovk

jmcouffin:
and I do happily criticism/ you got a good point @nasmovk thanks

PR #2791, opened 2025-09-02
CLI attach command --help appears to be outdated

frank-e-loftus:
CLI attach command --help appears to be outdated

The --help for the attach command gives the option "latest" which does not appear to be supported anymore:

image

Unless I'm misunderstanding, I believe it should instead list "DEFAULT".

Same for the "dynamosafe" option. It appears to be unsupported:

image


jmcouffin:
It looks good, thanks @frank-e-loftus

PR #2811, opened 2025-09-23
PY3 compatibility for itertools

### LINKED ISSUE #2810: [Bug]: output.print_table() won't work for IPY3
### ✈ Pre-Flight checks

### 🐞 Describe the bug

I think this is due to izip_longest being renamed to zip_longest in IPY3. I'll try to fix it after some testing

### ⌨ Error/Debug Message

```shell
File "C:\Program Files\pyRevit-Master\pyrevitlib\pyrevit\output\__init__.py", line 582, in print_table
AttributeError: 'module' object has no attribute 'izip_longest'
```

### ♻️ To Reproduce

_No response_

### ⏲️ Expected behavior

_No response_

```shell
master | Branch: "master" | Version: "5.2.0.25181+1332" | Path: "C:\Program Files\pyRevit-Master"
```

### Additional context

_No response_

Wurschdhaud:
## Description

Fixes itertools compatibility

---

## Checklist

  ```bash
  ```
- [ ] Changes are tested and verified to work as expected. **- only partially!**

---

## Related Issues

- Resolves #2810

---

## Additional Notes

The table won't show up in IPY3 unless i put a empty print statement before end of the print_table function - no idea why. IPY2 works fine. Like so:

```
        table = header + justifier + rows
        print("")
        self.print_md('### {title}'.format(title=title))
        self.print_md(table)
```

---

### DIFF
--- a/extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer Tools.panel/Generate API Stubs.pushbutton/script.py
+++ b/extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer Tools.panel/Generate API Stubs.pushbutton/script.py
@@ -7,26 +7,67 @@ from pyrevit.coreutils import assmutils
 from pyrevit import labs
 from pyrevit import forms
 from pyrevit import script
+from rpw.ui.forms.flexform import FlexForm, Label, Button, CheckBox
 
 
 logger = script.get_logger()
 output = script.get_output()
 
-
-ASSMS = [
+RVT_ASSMS = [
     'RevitAPI',
-    'RevitAPIUI',
+    'RevitAPIUI'
+]
+DYN_ASSMS = [
+    'DynamoServices',
+    'DynamoCore',
+    'DynamoApplications',
+    'DSCoreNodes',
+    'RevitServices',
+    'RevitNodes',
+    # 'ProtoGeometry',
     # 'UIFramework',
     # 'Adwindows',
+]
+PYR_ASSMS = [
     'pyRevitLabs.Common',
     'pyRevitLabs.TargetApps.Revit',
-    'pyRevitLabs.PyRevit',
+    'pyRevitLabs.PyRevit'
 ]
+
+def flexform_select():
+    components = [
+        Label('Select Assembly Group to Generate Stubs:'),
+        CheckBox('Revit', 'Revit Assemblies', default=True),
+        CheckBox('Dynamo', 'Dynamo Assemblies', default=False),
+        CheckBox('pyRevit', 'pyRevit Assemblies', default=False),
+        Button('Pick Destination Folder and Continue'),
+    ]
+    form = FlexForm('Generate API Stubs', components)
+    form.show()
+    selected = []
+    if form.values.get('Revit'):
+        selected.extend(RVT_ASSMS)
+    if form.values.get('Dynamo'):
+        selected.extend(DYN_ASSMS)
+    if form.values.get('pyRevit'):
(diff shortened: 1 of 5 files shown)


Wurschdhaud:
https://github.com/IronLanguages/ironpython3/wiki/Upgrading-from-IronPython2#redirecting-output

This might have sth to do with that weird print statement?

Wurschdhaud:
AI to the rescue:

Markdown not rendered in IPy 3.4 → Output buffering / stream behavior change (binary vs text writer)
print("") works → Forces stdout flush or stream initialization
Real fix → Add sys.stdout.flush() after printing in print_md()
Best fix if hosting IronPython → Configure ConsoleSupportLevel = SupportLevel.Basic in host code

I can confirm that this inside the output ini works:
```
import sys

@staticmethod
def print_md(md_str):
    ...
    print(html_code, end="")
    sys.stdout.flush()  # <<< Add this line
```
and i assume that changing dev>modules>pyReviLabs.ironpython3>Src>DLR>Src>Microsoft.Scripting>Runtime>ShareIO.cs
from SupportLevel.Full to SupportLevel.Basic might also solve this issue

But what are the downsides of either solution?

jmcouffin:
you could add and test adding 
```py
if PY3:
    sys.stdout.flush()
```
to the print_md function.
While I agree it may be smarter to fix it at the ShareIO level, it may be much more difficult to see the side effects, and would require extensive testing.

thoughts @sanzoghenzo @dosymep

sanzoghenzo:
while you're at it, can you drop the `zipper` variable altogether and just call `izip_longest` directly?

I don't see any benefit of using that support variable.

> thoughts @sanzoghenzo @dosymep

I would go with this

> Best fix if hosting IronPython → Configure ConsoleSupportLevel = SupportLevel.Basic in host code

since one of the goals I had in mind when we talked about isolating the plugin to avoid DLL hell was to drop our custom libraries and leverage the upstream ones via nuget (just to avoid having to remember what we patched when we need to update it) and use ilrepack or something like that. But I'm not the guru on this matter :sweat_smile:

sanzoghenzo:
On a more general note, I believe we should start to rely more and more on [six](https://github.com/benjaminp/six/tree/main). It's already in the `site-package`, but it may be worth to update it to 1.17.0, which is the latest release from december 2024, and will be the last one supporting python 3.3 through 3.5.

Wurschdhaud:
So I tried it with six, but i guess that site-package is loaded later? Which is a shame, would have made it more compact.

sanzoghenzo:
site-package should be added to the path by the ironpython executor, so six should be there...

What did you try? Does `import six` raise an error?

Wurschdhaud:
Yeah, i tried both:
`import six` for testing
as well as
`from six.moves import zip_longest`.

During startup i get the `ImportError: No module named six`, and pyrevit doesnt load at all.

if i add it later, and hit reload in the pyrevit tab, it works

sanzoghenzo:
Oh, I didn't think about pyrevitloader, it also loads pyrevit library.
Since is a single module, we could copy it directly inside pyrevitlib.

But I believe we can postpone it to another PR 😉

Wurschdhaud:
So should i drop this PR and wait for the pyrevitloader→pyrevitlib fix?

sanzoghenzo:
> So should i drop this PR and wait for the pyrevitloader→pyrevitlib fix?

Sorry for not being clear, the PR is OK as it is.

I was referring to introducing six, it should be done extensively on the entire codebase, so I believe it is out of scope of this PR and can be postponed.

Unless you want to fix this right away!

PR #2827, opened 2025-09-30
Rewrite of the print_table function: module "output" __init__.py

nasmovk:
## Rewrite of the print_table output method.

This is in response to the need for greater control over column widths and other formatting options such as borders, alignments, and row striping.

The method is rewritten to directly output an HTML table instead of the original markdown approach. It introduces more CSS options through classes. This in turn allows different formatting options to be specified for each table in the same output window (which was limited in the original).

The new functionality can be optionally called with new kwargs. If these are not specified, the original functionality of the old print_table is maintained. 

It adds some data validation to check that arguments of type list are of the same length.

Another improvement: It allows several tables in one output to have their own formatting options.

Finally, this contribution should make it easier to later implement HTML `colspan `options to make more flexible tables/

## Related Issues 
https://discourse.pyrevitlabs.io/t/pyrevit-output-table-formatting/1532/13

## Additional Notes
I attach a button script that runs the new method to show it in action.

Nobody asked for contribution, but it is very useful to me already.
I am sure my python is amateurish, but I think it works well.

[TABOUTPUTDEV_script.py](https://github.com/user-attachments/files/22623539/TABOUTPUTDEV_script.py)


pyrevitlib/pyrevit/output/__init__.py
```diff
@@ -548,19 +543,105 @@ def print_md(md_str):
...
+            
+        if columns and len_data_row != len(columns): # columns is allowed to be None
+            return False, "Column head list length not equal to data row"
+        
+        if formats and len_data_row != len(formats): # formats is allowed to be None
+            return False, "Formats list length not equal to data row"
+        
+        # Next check **kwargs
+        # Loop through the lists and return if not a list or len not equal
+        for l in input_kwargs:
```
jmcouffin: The return True, "kwarg None" should be continue to ensure all kwargs are validated.

```
for l in input_kwargs:
    if not l: # No kwarg is OK beacause they are optional
        continue
    if not isinstance(l, list):
        # ...
```

pyrevitlib/pyrevit/output/__init__.py
```diff
@@ -573,56 +654,127 @@ def print_table(self, table_data, columns=None, formats=None,
...
+        html += "<tbody>"
+        for row in table_data:
+            # Build an HTML data row with CSS class for this table
+            html += "<tr class='data-row-{}'>".format(table_uid)
+            for i, cell_data in enumerate(row):
+                
+                # Slight workaround to be backwards compatible with pyRevit original print_table
+                # pyRevit documentation gives the example: formats=['', '', '', '{}%'],
+                # Sometimes giving an empty string, sometimes a placeholder with string formatting
+                if formats: # If format options provided
```
jmcouffin, line 829: less verbose
```
if formats:
    format_placeholder = formats[i] if formats[i] else "{}"
    cell_data = format_placeholder.format(cell_data)
```

pyrevitlib/pyrevit/output/__init__.py
```diff
@@ -573,56 +654,127 @@ def print_table(self, table_data, columns=None, formats=None,
...
+        column_widths                = kwargs.get("column_widths", None)
+        column_vertical_border_style = kwargs.get("column_vertical_border_style", None)
+        table_width_style            = kwargs.get("table_width_style", None)
+        repeat_head_as_foot          = kwargs.get("repeat_head_as_foot", False)
+        row_striping                 = kwargs.get("row_striping", True)
+
+
+        # Set a unique ID for each table
+        # This is used in HTML tags to define CSS classes for formatting per table
+        table_uid = random.randint(10000, 99999)
```
jmcouffin: better use a counter, there is a small chance you hit a bug 
set the variable somewhere higher:
`self._table_counter = 0`
then 
```
self._table_counter += 1
table_uid = self._table_counter
```

jmcouffin (changes requested):

jmcouffin:
Have you tested it on tools containing old table behaviour: preflight checks audit all, line counter for example

jmcouffin:
if it works well, this is a ool change that could require documentation on the notion docs.

nasmovk:
Hi Jean-Marc,
Thanks for the feedback.
I will make changes according to your comments.
Yes, I have tested it on all the existing pyRevit tool buttons that call the original print_table. I will recheck after making the next changes.

nasmovk:
I have made a new Commit with the suggested changes.
I confirm I have tested on the existing tools that use print_table.
I agree that the Notion docs should be updated if my changes are accepted. I am happy to help with that.

jmcouffin:
> I agree that the Notion docs should be updated if my changes are accepted. I am happy to help with that.

Can you take the content of the said page and make a new one on your own notion account, share me the link to update. 
We don't have a pro account to share edit rights with Ehsan outside of the both of us. DM me ht elink in the forum

nasmovk:
Hi. Yes, I'll get started on the Notion page. 
Thanks.

jmcouffin:
> Hi. Yes, I'll get started on the Notion page.
> Thanks.

I made the change to the docs. 
Thansk @nasmovk

PR #2857, opened 2025-10-09
Update report.py - Remove Circular referencing between report.py and …

tay0thman:
…script.py

This issue was discovered while the preflight checks were throwing exceptions, after further investigation, it seems that the issue was caused by circular chain referencing of script.py, while script.py references genericcomps.GenericUICommand which itself references script.py again.

The solution is to remove script.py from the imports of report.py

# Update report.py - Remove Circular referencing between report.py and …

##This issue was discovered while the preflight checks were throwing exceptions, after further investigation, it seems that the issue was caused by circular chain referencing of script.py, while script.py references genericcomps.GenericUICommand which itself references script.py again.
---

## Checklist

- [ ok] Code follows the [PEP 8](https://peps.python.org/pep-0008/) style guide.
- [ok ] Code has been formatted with [Black](https://github.com/psf/black) using the command:
  ```bash
  ```
- [ ] Changes are tested and verified to work as expected.

---

## Related Issues

- Resolves #[issue number]

---

## Additional Notes

---

### DIFF
--- a/pyrevitlib/pyrevit/revit/report.py
+++ b/pyrevitlib/pyrevit/revit/report.py
@@ -1,6 +1,6 @@
 """"Utility methods for reporting Revit data uniformly."""
 
-from pyrevit import DB, script, HOST_APP
+from pyrevit import DB, HOST_APP
 from pyrevit.output import PyRevitOutputWindow
 from pyrevit.revit import query
 


jmcouffin:
Good one. I spotted the issue yesterday and could not identify the cause quickly enough.
Thanks @tay0thman

tay0thman:
> Good one. I spotted the issue yesterday and could not identify the cause quickly enough. Thanks @tay0thman

Preflight checks are so critical, can't sleep without making them working.

jmcouffin:
> > Good one. I spotted the issue yesterday and could not identify the cause quickly enough. Thanks @tay0thman
> 
> Preflight checks are so critical, can't sleep without making them working.

Glad you live for it. This is the tool I wanted to build forever and that got me to really code and learn to code harder

PR #2860, opened 2025-10-13
Modifying the Parameter Values to Parameter Script: Adding separator and modifying target parameters

ramyamaher:
Modifying Parameter Values to Parameter Script

1. Adding a separator character choice to loop from instead of " - "
2. Adding the ability to add spaces before or after / or before and after the separator
3. Target parameters shown in the target parameter list are only text parameters that can be overridden


jmcouffin:
@ramyamaher 
Next time, make it a draft PR until this is ready to be commented.
Good list of enhencement!

PR #2869, opened 2025-10-15
tool: Measure 3D

Wurschdhaud:
## Description

Since the in Revit 2023ish introduced measuring tool isn't really that helpful, this would be my shot at it.
To have cross-version compat, visual aids are created as directshapes, not loaded as families.
Some variables are configurable - if a config file makes sense. Could also imagine disabling the visual aids via config?

Feedback welcome.

image

---

## Checklist

  ```bash
  ```

---

## Related Issues

- Resolves #[issue number]

---

## Additional Notes

Translations AI generated.
Code supported by AI - feedback welcome if it isn't pretty enough for the repo.
Template for modeless: 
https://github.com/CyrilWaechter/pyRevitMEP/tree/b7036956cf0a47c510b9f61711d4c86224d50e40/pyRevitMEP.tab/Samples.panel/Samples.pulldown/FormExternalEventHandler.pushbutton

---

### DIFF
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/3D.pulldown/Measure.pushbutton/script.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/3D.pulldown/Measure.pushbutton/script.py
@@ -10,8 +10,8 @@ logger = script.get_logger()
 # Document variables
 doc = revit.doc
 uidoc = revit.uidoc
-distance_unit = doc.GetUnits().GetFormatOptions(DB.SpecTypeId.Distance).GetUnitTypeId()
-unit_label = DB.LabelUtils.GetLabelForUnit(distance_unit)
+length_unit = doc.GetUnits().GetFormatOptions(DB.SpecTypeId.Length).GetUnitTypeId()
+unit_label = DB.LabelUtils.GetLabelForUnit(length_unit)
 
 # Global variables
 measure_window = None
@@ -51,16 +51,16 @@ def calculate_distances(point1, point2):
 def format_distance(value_in_feet):
     return DB.UnitFormatUtils.Format(
         doc.GetUnits(),
-        DB.SpecTypeId.Distance,
+        DB.SpecTypeId.Length,
         value_in_feet,
         False,
     )
 
 
 def format_point(point):
-    x = DB.UnitUtils.ConvertFromInternalUnits(point.X, distance_unit)
-    y = DB.UnitUtils.ConvertFromInternalUnits(point.Y, distance_unit)
-    z = DB.UnitUtils.ConvertFromInternalUnits(point.Z, distance_unit)
+    x = DB.UnitUtils.ConvertFromInternalUnits(point.X, length_unit)
+    y = DB.UnitUtils.ConvertFromInternalUnits(point.Y, length_unit)
+    z = DB.UnitUtils.ConvertFromInternalUnits(point.Z, length_unit)
     return "({:.2f}, {:.2f}, {:.2f})".format(x, y, z)
 
 


tay0thman:
I love this idea!
I quickly skimmed through the script @Wurschdhaud . have you considered using directcontext3d instead of creating elements within a transaction? I think it is better approach to experiment with.
https://www.autodesk.com/autodesk-university/class/DirectContext3D-API-Displaying-External-Graphics-Revit-2017

Wurschdhaud:
At first glance that seems to be even better. Haven't heard of it before, I'll have to dive into it.

EDIT:
i was thinking it would be cool if you could move the visual aid spheres and the measurement updates dynamically. Don't know if this will fit with the directcontext3d thing, at least not for the spheres.

tay0thman:
@Wurschdhaud here is a sample implementation of DC3D in our Dev tools .

https://github.com/pyrevitlabs/pyRevit/tree/develop/extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer%20Examples.panel/DirectContext3D.pushbutton

tay0thman:
another implementation example with the find view range :
https://github.com/pyrevitlabs/pyRevit/blob/develop/extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/toggles3.stack/ViewRange.pushbutton/script.py

Wurschdhaud:
Added! I really like your idea with DC3D. This unfortunately means I have to redo a lot of tools in my company to also use it :)

extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/3D.pulldown/Measure.pushbutton/script.py
```diff
@@ -0,0 +1,288 @@
...
+        with forms.WarningBar(title="Pick second point"):
+            point2 = revit.pick_elementpoint(world=True)
+            if not point2:
+                return
+
+        new_meshes = create_measurement_meshes(point1, point2)
+
+        # Add to existing meshes (don't replace)
+        if dc3d_server:
+            existing_meshes = dc3d_server.meshes if dc3d_server.meshes else []
```
sanzoghenzo:
```suggestion
            existing_meshes = dc3d_server.meshes or []
```

extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/3D.pulldown/Measure.pushbutton/script.py
```diff
@@ -0,0 +1,288 @@
...
+    if point1.DistanceTo(point2) > 0.001:
+        meshes.append(create_line_mesh(point1, point2, LINE_COLOR_DIAG))
+
+    return meshes
+
+
+def delete_all_visual_aids():
+    """Delete all visual aids by clearing the DC3D server meshes."""
+    global dc3d_server
+
```
sanzoghenzo: since you don't re-assign the variable, the `global` keyword is not needed.
```suggestion
```

extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/3D.pulldown/Measure.pushbutton/script.py
```diff
@@ -0,0 +1,288 @@
...
+            format_distance(dx),
+            format_distance(dy),
+            format_distance(dz),
+            format_distance(diagonal),
+        )
+        measurement_history.append(history_entry)
+
+        # Keep only last MAX_HISTORY measurements
+        if len(measurement_history) > MAX_HISTORY:
+            measurement_history = measurement_history[-MAX_HISTORY:]
```
sanzoghenzo: instead of using a list, you can use a [deque](https://docs.python.org/2/library/collections.html#collections.deque) with `maxlen=MAX_HISTORY`.

You can then just `.append` without having to check the length of the collection and drop the `global` at the beginning of this function since you don't re-assign it.

Wurschdhaud:
Thanks for the feedback with explanation, really appreciated @sanzoghenzo !

thumDer:
> Added! I really like your idea with DC3D. This unfortunately means I have to redo a lot of tools in my company to also use it :)

Glad you've found it useful!

jmcouffin:
@Wurschdhaud 
Cool stuff again, 
image
- the window displays the correct project units label bu the conversion still show internal units in decimal feet.
- don't show the label for units next to the points coordinates

Wurschdhaud:
thanks for the feedback!
Changed the unit to the correct type "distance", and used the other format tool.

jmcouffin:
Not there yet: 
image

- no units on the XYZ ✅ 
- units and conversion on length/distance 🟥

Wurschdhaud:
It uses "Distance" project settings, as it is a distance measurement - if that's set to unitless, it stays unitless. Is it set to unitless in your project @jmcouffin ?
That way it's no hard coded value. But can be confusing i guess?

image

jmcouffin:
@Wurschdhaud set to feet in my project.
and there is a rounding happening if using feet (distance)
It is better to use length IMO

Wurschdhaud:
Yeah, rounding (blue) and unit (red) are set in via UI of distance (green) in the project units. I'd see that as a benefit, since UI is more accessible and flexibel.

image

jmcouffin:
yeah, but default to no decimal, and for imperial system, this is a killer.
Length should be the default

Wurschdhaud:
Sounds like you need to update your project templates :wink:
I still think as we're measuring a distance, the correct unit would be distance, not length, but i can change it

jmcouffin:
> Sounds like you need to update your project templates 😉

Never



> distance

It cannot be set to feet and fractional inches _ no go

Wurschdhaud:
- distance calculation output format according to project unit interface
- xyz points hardcoded to length unit with rounding for 2 digits as before, no unit displayed

image

jmcouffin:
Great @Wurschdhaud You are starting to have quite a lot of code in the code base this month! Thanks for that.

PR #2883, opened 2025-10-23
Add Relink Textures tool to pyRevitTools

Swichllc:
## New Tool:  Relink Textures

**Description:** Automates finding and relinking missing texture paths in Revit materials

**Location:** pyRevit tab > Modify panel

**Features:**
- Automatically finds missing textures
- Relinks paths in seconds
- Search for textures not just location, in case someone moved the files

**Testing:** Tested on Revit 2025

### DIFF
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/edit3.stack/Edit.pulldown/Relink Textures.pushbutton/script.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/edit3.stack/Edit.pulldown/Relink Textures.pushbutton/script.py
@@ -4,6 +4,7 @@ from os import walk
 from os.path import join, isdir, isfile, basename
 import traceback
 import collections
+import fnmatch
 
 from pyrevit import HOST_APP, forms, script, revit, EXEC_PARAMS
 from pyrevit import DB
@@ -126,13 +127,40 @@ class TextureIndexer:
         self.index = collections.defaultdict(list)
         valid_roots = self.unique_existing_paths(roots)
         
+        # Define common texture file extensions
+        texture_extensions = ('*.jpg', '*.jpeg', '*.png', '*.bmp', '*.tga', '*.tiff', '*.tif', '*.dds')
+        
+        # Directories to skip for faster traversal
+        skip_dirs = {
+            # Version control
+            '.git', '.svn', '.hg',
+            # Build/cache
+            'node_modules', '__pycache__', '.pyc',
+            # System files
+            '.DS_Store', 'Thumbs.db', 'Desktop.ini',
+            # Windows system
+            '$RECYCLE.BIN', 'System Volume Information',
+            # Revit specific
+            'backup', 'Backup'
+        }
+        
         output.log_info("Indexing {} folder(s)...".format(len(valid_roots)))
         
         for i, root in enumerate(valid_roots):
             try:
-                for dirpath, _, files in walk(root):
-                    for filename in files:
-                        self.index[filename.lower()].append(join(dirpath, filename))                    
+                for dirpath, dirs, files in walk(root):
+                    # Remove directories we want to skip (including all hidden folders)
+                    dirs[:] = [d for d in dirs if d not in skip_dirs and not d.startswith('.')]
+                    
+                    # Filter files by texture extensions using fnmatch
+                    texture_files = []
+                    for pattern in texture_extensions:
+                        texture_files.extend(fnmatch.filter(files, pattern))
+                    
+                    # Add texture files to index
+                    for filename in texture_files:
+                        self.index[filename.lower()].append(join(dirpath, filename))
+                        
             except (OSError, IOError) as error:
                 logger.warning("Failed to index {}: {}".format(root, error))
         


extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Relink Textures.pushbutton/script.py
```diff
@@ -0,0 +1,401 @@
+# -*- coding: utf-8 -*-
+"""Relink material textures for Revit via pyRevit."""
+
+__title__ = "Relink\nTextures"
```
tay0thman: These are duplicate with YAML, not needed

extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/edit3.stack/Edit.pulldown/Relink Textures.pushbutton/icon.png
tay0thman, line 1: This icon is not consistent with pyrevit Icons, please replace with something that is more visually consistent

tay0thman:
@Swichllc Added few comments on the structure of the script, I will fetch the PR and test it locally before running code review.

extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/edit3.stack/Edit.pulldown/Relink Textures.pushbutton/icon.png
tay0thman: also please consider dark+light UI

tay0thman:
also please propose a different location, I would suggest having it under the project/ links pulldown button

This is not an everyday tool so it should not take a full stack.

Swichllc:
@tay0thman 
I'll work on these comments, as I mentioned in the forum, before I got the green light to create PR, Im very new to all of this and Im using AI to build it, so it could use a review. 
**For the dark/light mode - besides adding "icon_dark.png" I couldn't find the answer on how to force dark/light mode on the icon, that solution didn't work when I tested it with a new icon that match the pyRevit icons. 
**For the location - a few thoughts - Links is a bit confusing since its a very specific term and flow, same with Family  - right now I don't see anything relevant, so I added it under Project, since its a general action needed to do once per project. 
From a users' perspective, if the previous placement is an issue, What I suggest is to create a new pulldown menu Project/Textures so its not a large standalone, and it can include more texture related commands - I have a few more tools ideas in that area.

jmcouffin:
> **For the dark/light mode - besides adding "icon_dark.png" I couldn't find the answer on how to force dark/light mode on the icon, that solution didn't work when I tested it with a new icon that match the pyRevit icons.

use the icons8 website to find one.
download to 96px png
icon.png for the regular UI
icon.dark.png for the Dark theme

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Relink Textures.pushbutton/script.py
```diff
@@ -0,0 +1,401 @@
...
+                logger.error("Asset edit failed for {}: {}".format(a_elem.Name, ex))
+            finally:
+                if scope:
+                    try:
+                        scope.Dispose()
+                    except:
+                        pass
+        
+        t.Commit()
+        logger.info("Transaction committed successfully")
```
Wurschdhaud: more pyrevit like would be 
```python
with revit.Transaction("do stuff"):
```

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Relink Textures.pushbutton/script.py
```diff
@@ -0,0 +1,401 @@
...
+try:
+    # Check if SHIFT is pressed for quick-config
+    shift_click = False
+    try:
+        shift_click = __shiftclick__
+    except:
+        pass
+    
+    if shift_click:
+        run_config()
```
Wurschdhaud: i think you can skip the try. Also according to notion you should use `EXEC_PARAMS.config_mode`

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Relink Textures.pushbutton/script.py
```diff
@@ -0,0 +1,401 @@
...
+# Document and logger
+doc = __revit__.ActiveUIDocument.Document
+logger = script.get_logger()
+output = script.get_output()
+
+# Configuration file path
+CONFIG_FILE = os.path.join(
+    os.getenv('APPDATA'),
+    'pyRevit_TextureRelink_Config.json'
+)
```
Wurschdhaud: have you considered `script.get_config()` ?

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Relink Textures.pushbutton/script.py
```diff
@@ -0,0 +1,401 @@
...
+    Material, 
+    AppearanceAssetElement,
+    ElementId, 
+    Transaction
+)
+from Autodesk.Revit.DB.Visual import (
+    AppearanceAssetEditScope, 
+    UnifiedBitmap, 
+    Asset, 
+    AssetPropertyString
```
Wurschdhaud: more pyrevit like would be from `pyrevit import DB` and work with that
but this is fine imo

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Relink Textures.pushbutton/script.py
tay0thman: also wrapping large chunks of code using try / except is highly risky, the logic should be controlled as much as possible

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Relink Textures.pushbutton/script.py
```diff
@@ -0,0 +1,401 @@
+# -*- coding: utf-8 -*-
+"""Relink material textures for Revit via pyRevit."""
+
+__title__ = "Relink\nTextures"
+__author__ = "SwichTools"
+
+# Standard library imports
+import os
```
tay0thman: unless the entire os library is needed, I would prefer to import the path class only,  this will reduce the memory footprint.
import os.path as op

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Relink Textures.pushbutton/script.py
```diff
@@ -0,0 +1,401 @@
...
+)
+from Autodesk.Revit.DB.Visual import (
+    AppearanceAssetEditScope, 
+    UnifiedBitmap, 
+    Asset, 
+    AssetPropertyString
+)
+
+# Document and logger
+doc = __revit__.ActiveUIDocument.Document
```
tay0thman: simpler for pyrevit

doc = revit.doc

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Relink Textures.pushbutton/script.py
```diff
@@ -0,0 +1,401 @@
...
+__author__ = "SwichTools"
+
+# Standard library imports
+import os
+import json
+import traceback
+import collections
+
+# pyRevit imports
+from pyrevit import forms, script
```
tay0thman: from pyrevit import forms, script, revit, DB

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Relink Textures.pushbutton/script.py
```diff
@@ -0,0 +1,401 @@
...
+        
+        t.Commit()
+        logger.info("Transaction committed successfully")
+        
+    except Exception as e:
+        logger.error("Relink transaction failed: {}".format(e))
+        logger.error(traceback.format_exc())
+        
+        try:
+            t.RollBack()
```
tay0thman: also wrapping parts if transaction elements with try, will likely cause errors down the line

extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/edit3.stack/Edit.pulldown/Relink Textures.pushbutton/script.py
```diff
@@ -0,0 +1,401 @@
...
+        )
+        return
+
+    # Show results
+    msg = "Appearance assets examined: {}\nTextures relinked: {}".format(examined, fixed_count)
+    
+    if unresolved:
+        msg += "\n\nUnresolved textures: {} (see output panel for list)".format(len(unresolved))
+        logger.warning("=" * 60)
+        logger.warning("UNRESOLVED TEXTURE NAMES:")
```
tay0thman: I see an excessive amount of using logger. It should be minimized. larger amounts of logging will add further CPU cycles.

extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/Relink Textures.pushbutton/script.py
Wurschdhaud: So strictly speaking: would this be even more efficient? Although I guess negligible?

`from pyrevit.framework import op
`

tay0thman: It is negligible considering that you have an open revit model that uses 6GB of RAM. The point here is to emphasize good coding practices,  since people are looking at these scripts to learn . So it is better to set a good example for them.

jmcouffin:
@Swichllc are you comfortable enough to make all these changes?

Swichllc:
@jmcouffin yes, I'm very comfortable to try fixing it. 
But I'll update you if I come across a problem I can't solve. 
Thank you

extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/edit3.stack/Edit.pulldown/Relink Textures.pushbutton/icon.png
Swichllc: image
image

icon
icon_dark

It seems the dark/light is a deeper issue, and not related to adding the 2 types of icons - none of the pyRevit icons switch on my computer when changing the theme - I think its a local issue on my system, so my 2 new icons that I will upload with the update should work

Swichllc:
@jmcouffin @tay0thman @Wurschdhaud 
1. I have attempted to make the changes - some of them caused the search to crash, and I couldn't pinpoint which one so I pivoted. 
2. Scanning an entire drive (c:/ for example) still works, it is not as fast as it was with the original code, but it still works, even though that is the extreme case, I do believe it necessary to make sure it possible. 
3. Location is still the same - besides creating a new Textures dropdown in Project.Panel, and add more texture related tools in the future, I still think its a standalone function that would be used once per project, if you guys think it should be somewhere else I will not object. 
4. At this point if there are crucial code issue, since Im a newbie, any help would be appreciated.
5. Icons are png 96x96 with dark and light versions
Thank you

jmcouffin:
> 3\. Location is still the same - besides creating a new Textures dropdown in Project.Panel, and add more texture related tools in the future, I still think its a standalone function that would be used once per project, if you guys think it should be somewhere else I will not object.

I would put it under the **modify** panel in the **edit** drowpdown

> 4\. At this point if there are crucial code issue, since Im a newbie, any help would be appreciated.

Will review and polish this week if needed

extensions/pyRevitTools.extension/pyRevit.tab/Modify.panel/edit3.stack/Edit.pulldown/Relink Textures.pushbutton/icon.png
tay0thman: It is not a local issue, you need to restart revit to get some of the pyrevit Icons fully updated.

jmcouffin:
I went through, did a lot of cleaning, reorganizing, some with cursor, some without.
It seems to work smoothly. I tried to go as granular as I could with the time I had. 
Would you mind testing it @Swichllc and @tay0thman ?

Swichllc:
@jmcouffin 
image
Screenshot 2025-10-27 193752

image

I tested it and it seems to miss the mark, it didnt find or relink any texture, as you can see in the screenshot.
I also didnt get the report of which textures could not be found.
I ran the original code with the same Revit project, same folder setup, and you can see the amount of textures it found and relink.

jmcouffin:
Reviewed the user flow, simplified some checks and properly tested it this time.
@Swichllc this should improve the whole user experience.
If you are happy with it as well, I will merge this PR

Swichllc:
@jmcouffin tested, its working - thank you for helping me with this
It is missing the completion report (attached image), but it does give the error report 
Im ok without it right now
image

jmcouffin:
@Swichllc I felt the number of assets examined (and counted) was not a crucial info. I left the "Textures relinked" in the output window + unresolved list as well.

jmcouffin:
Thanks for this tool. This is a great addition

Swichllc:
@jmcouffin Perfect, Thank you again

PR #2915, opened 2025-11-08
chore: tools Minor fixes

Wurschdhaud:
## Description

The type of minor annoyances you only notice after a few weeks time:
- lib commands used when possible
- accidentally used global variable
- add formatting and slope to measure 3D tool

---

## Checklist

- [X] Code has been formatted with black


jmcouffin:
Great stuff,
If you don't mind, next time, try to break down a PR like this one in two: one for changing the usage of pyrevit lib specific method, another to add slope to 3d measurements tools

PR #3007, opened 2026-01-10
Load-family-fix

jmcouffin:
## Description

tidy up behavor and document new return of load_family def

- ref #3006


jmcouffin:
@devloai do not review this PR

PR #3030, opened 2026-01-23
Update Create Workset For Linked Element

Denver-22:
Update Create Workset For Linked Element:
- Added a prompt to indicate whether collaboration should be enabled in the document.
- All Worksets are created in a single transaction.
- Removed "ZL_" from the Workset name prefixes for RVT and DWG.


jmcouffin:
Thank you @Denver-22, a no brainer.

I would have kept the ZL (a convention I used to keep all the links worksets at the end)

Denver-22:
@jmcouffin, I didn't know that was a convention.
I'll bring back "ZL_" in the next PR.

Issue #2519, opened 2025-01-14
RFC: Installers overhaul


sanzoghenzo:
After long thinking and research for possible solutions, I would like to propose a cleanup of the installers situation.
What I would like to see:

- A single installer with all users/current user selection, so that there's no confusion on what installer to download
- potentially a single installer for both cli and full version with component selection - IT admins uses the CLI because is the only documented thing, but they usually don't go through the process of customizing the clones, so the time gained by downloading a smaller installer is totally lost by cloning the repo.

We could get rid of the .MSI and .nuget packages since the Inno Setup exe has [CLI options to personalize the installation](https://jrsoftware.org/ishelp/index.php?topic=setupcmdline) and the [ability to run with or without admin](https://jrsoftware.org/ishelp/index.php?topic=setupcmdline)
Obviously we need to document the aforementioned inno setup options and the valid /components values on the wiki, and put a link of that page in the release message.

Alternatively, we can drop inno setup and use WiX Toolkit to create only the MSI package.
This could solve the problems in some corporate environment where the execution of .exe outside specific folders is prohibited.
But for this I/we have to learn how to use WiX and see if has all the feature we need.

Do you see any problem in any of this scenarios?
ccin @jmcouffin @dosymep and @eirannejad , but anyone is welcome to bring insight to the issue!

As a side note, I believe there's an error in the CI/`pyrevit set products` command: installers GUID must stay the same in order to detect versions of the same app.

jmcouffin:
- single installer ... ,💯
- having the cli packaged with pyrevit itself. While I understand and agree, I am a bit worried we lose some people in the process
- exe have been asked in the past, but this was far less than MSI, just because of the lock on exe installers in some environments like you mentioned. All in for MSI and WIX... It would be better if we all learn and document in the process
- 'set products' I don't know if this is by design: to be able to install multiple versions of pyrevit (4.8 and 5.0 for example). Day dreaming here, If the WIX installer could offer an option to install wip and release and also take care of a proper cleanup when uninstalled...

Sure you don't want to stay @sanzoghenzo 🙏😨

dosymep:
WIX it's a simple tool and it's quite possible to make an installer on it
I think it would be a better solution to abandon the old installer and make a single installer in C#

sanzoghenzo:
Alright then, wix it is... and it makes sense to use a tool that uses dotnet build, so that we need one less thing in the ci toolchain (this also applies to nuke as discussed before!)

[...]

PR #3058, opened 2026-02-01
Fix Inno Setup installer GUIDs (RFC #2519)

jmcouffin:
## Issue

Addresses **RFC #2519**: "installers GUID must stay the same in order to detect versions of the same app."

Previously, `set_product_data` (run via `pyrevit set products` in CI) generated a **new UUID** for pyRevit and pyRevit CLI on every run and wrote it into the Inno Setup scripts (`.iss` files) as `MyAppUUID`. As a result:

- Each build was treated by Windows as a **different application**
- Multiple entries for "pyRevit" and "pyRevit CLI" appeared in Add/Remove Programs
- Upgrade (install new version over old) and uninstall behavior were broken

## Change

**File:** `dev/_release.py`

In `set_product_data`, the two calls that overwrote the Inno Setup AppId were **removed**:

- `_installer_set_uuid(pyrevit_pc, configs.PYREVIT_INSTALLER_FILES)`
- `_installer_set_uuid(pyrevitcli_pc, configs.PYREVIT_CLI_INSTALLER_FILES)`

The `.iss` files already define stable GUIDs; they are no longer overwritten on each build:

- **pyRevit** (pyrevit.iss, pyrevit-admin.iss): `f2a3da53-6f34-41d5-abbd-389ffa7f4d5f`
- **pyRevit CLI** (pyrevit-cli.iss, pyrevit-cli-admin.iss): `9557b432-cf79-4ece-91cf-b8f996c88b47`

**Unchanged:**

- MSI installer UUID handling (`_msi_set_uuid`) is unchanged
- Product data file updates (`_update_product_data_file`) are unchanged
- The `_installer_set_uuid` helper remains in the codebase for possible future use

## Result

- Inno Setup AppIds stay constant across versions
- Windows sees one application per product; Add/Remove Programs shows a single entry for pyRevit and one for pyRevit CLI
- Upgrading to a new build replaces the previous install instead of creating a duplicate entry

### DIFF
--- a/dev/_release.py
+++ b/dev/_release.py
@@ -120,10 +120,6 @@ def set_product_data(_: Dict[str, str]):
     pyrevit_pc = _get_new_product_code()
     pyrevitcli_pc = _get_new_product_code()
 
-    # update product info on installer files
-    _installer_set_uuid(pyrevit_pc, configs.PYREVIT_INSTALLER_FILES)
-    _installer_set_uuid(pyrevitcli_pc, configs.PYREVIT_CLI_INSTALLER_FILES)
-
     # update product info on MSI installer files
     _msi_set_uuid(
         pyrevit_pc,


jmcouffin:
@devloai fix the critical issue in comment # 1 - Update the code to use stable GUIDs for the product database
@devloai add constants for the stable GUIDs to configs.py - Implement the suggested configuration constants

PR #3086, opened 2026-02-07
New helper for external events

Wurschdhaud:
### Goal
Add a helper function `execute_in_revit_context()` to simplify executing functions from modeless dialogs or non-Revit threads in the Revit API context.

### Problem
Currently, developers need to manually create `IExternalEventHandler` classes and `ExternalEvent` instances every time they want to call Revit API from modeless forms. This involves repetitive boilerplate code.

### Solution
Provides a simple one-liner function that handles the ExternalEvent setup internally, reducing boilerplate and improving code readability.

**Before:**
```python
class SimpleEventHandler(UI.IExternalEventHandler):
    def __init__(self, do_this):
        self.do_this = do_this
    def Execute(self, uiapp):
        try:
            self.do_this()
        except Exception as ex:
            logger.error("Error: {}".format(ex))
    def GetName(self):
        return "Event Handler"

handler = SimpleEventHandler(perform_measurement)
ext_event = UI.ExternalEvent.Create(handler)
ext_event.Raise()
```

**After:**
```python
execute_in_revit_context(perform_measurement, arg1, arg2, kwarg1=value1)
```

---

## Checklist

  ```bash
  ```

---

## Related Issues

- Resolves #[issue number]

---

## Additional Notes

- a developer example has been added
- the new function has been added to 3d measure
- fixed a bug with the new translation in 3d measure ("\n" would no longer do a new line)
- the new function has been added to section box
- the new function has been added to show view range

I tried to add it to color splasher as well, but that code is too much of an AI mess, will take longer. 

---

### DIFF
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/toggles3.stack/ViewRange.pushbutton/script.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Toggles.panel/toggles3.stack/ViewRange.pushbutton/script.py
@@ -79,16 +79,9 @@ class Context(object):
             )
             return False
 
-        try:
-            events.execute_in_revit_context(
-                self._update_view_range_internal, new_values, new_levels
-            )
-        except Exception as e:
-            if hasattr(self, "view_model"):
-                self.view_model.warning_message = (
-                    "Error executing view range update: {}".format(str(e))
-                )
-
+        events.execute_in_revit_context(
+            self._update_view_range_internal, new_values, new_levels
+        )
         return True
 
     def _update_view_range_internal(self, new_values, new_levels=None):


Wurschdhaud:
> `execute_in_revit_context` uses a single shared module-level handler/event and mutates `_HANDLER.func/args/kwargs` before raising. If multiple callers raise the event before Revit processes it (or from different threads), later calls can overwrite earlier ones and execute with the wrong args (or drop calls entirely). Consider making this thread-safe (lock around mutation + a queue of call requests drained in `Execute`), or create a dedicated handler/event per caller.

Is this actually an issue? I tried to force it by having multiple modeless windows open and hitting buttons - was working fine.

PR #3125, opened 2026-02-22
switch __shiftclick__ to EXEC_PARAMS.config_mode

Wurschdhaud:
## Description

As per https://pyrevitlabs.notion.site/Python-Runtime-Variables-1ae28b25faca48ff8c291d68e8010541 `__shiftclick__` should be avoided.
This PR moves all tools in pyRevit.tab to the exec_params approach to set a good example and make code consistent.

---

## Checklist

  ```bash
  ```

---

## Additional Notes

I did not test all tools, only random samples.

### DIFF
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Drawing Set.panel/Sheets.pulldown/List TitleBlocks on Sheets.pushbutton/script.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Drawing Set.panel/Sheets.pulldown/List TitleBlocks on Sheets.pushbutton/script.py
@@ -6,47 +6,7 @@ Select titleblocks on currently selected sheets in project browser.
 
 """
 # pylint: disable=import-error,invalid-name,broad-except,superfluous-parens
-from pyrevit import revit
-from pyrevit import forms
-from pyrevit import script
-
-output = script.get_output()
-logger = script.get_logger()
-
-
-def get_source_sheets():
-    sheet_elements = forms.select_sheets(
-        button_name="List TitleBlocks",
-        use_selection=True,
-        include_placeholder=False,
-    )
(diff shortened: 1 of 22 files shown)


extensions/pyRevitTools.extension/pyRevit.tab/Drawing Set.panel/Sheets.pulldown/List TitleBlocks on Sheets.pushbutton/script.py
```diff
@@ -6,47 +6,7 @@
 
 """
 # pylint: disable=import-error,invalid-name,broad-except,superfluous-parens
-from pyrevit import revit
-from pyrevit import forms
-from pyrevit import script
-
```
jmcouffin, line 12: Any reason to remove this big chunk?

jmcouffin:
This may be a breaking change for some external extensions.

You may want to change the definition of the dunder shift click and throw an explicit exception for the depreciation

extensions/pyRevitTools.extension/pyRevit.tab/Drawing Set.panel/Sheets.pulldown/List TitleBlocks on Sheets.pushbutton/script.py
Wurschdhaud: Yes, that portion was in the script twice

Wurschdhaud:
> This may be a breaking change for some external extensions.
> 
> You may want to change the definition of the dunder shift click and throw an explicit exception for the depreciation

You mean if i'd ditch the support entirely within the core? True, but i only changed it in the tools themselves to "set a good example". I think that's one of those legacy things that will get carried around forever :laughing:

PR #3126, opened 2026-02-22
Convert CAD Import to FreeForm/DirectShape – Robustness Improvements

Wurschdhaud:
## Description

Finishes up the long-standing beta status of this tool with several stability and UX fixes:

- Fixed crash when nothing is selected by falling back to pick_element(), with null and type checks on the result
- Fixed pick_element return type mismatch (selection.first was being called on a bare element)
- Filters selection to ImportInstance only, warns if multiple CAD imports are selected
- Moved DirectShape creation out of a per-mesh transaction into a single batched transaction
- Added zero-result warning when no solids or meshes are found in the import
- Fixed typo: "FreeFrom" → "FreeForm" in transaction name
- Added completion log message with solid/mesh counts

---

## Checklist

  ```bash
  ```


extensions/pyRevitTools.extension/pyRevit.tab/Project.panel/ptools.stack/Family.pulldown/Convert Imported 3D CAD.pushbutton/bundle.yaml
```diff
@@ -14,4 +14,4 @@ tooltip:
   es_es: Convertir símbolo CAD importado a elementos de forma libre o DirectShape
   de_de: Konvertiert importierte CAD Objekte in FreeForm oder DirectShape Elemente
   pt_br: Converter Símbolo CAD Importado para Elementos FreeForm ou DirectShape
-is_beta: true
+is_beta:
```
jmcouffin, line 17: @Wurschdhaud why removing the beta flag?

jmcouffin: I don't mind, I just want to know if it was on purpose. Leaving it empty might break the bundle parsing

Wurschdhaud: It's a bundle parsing test function
Just out of habit I guess, I tend to leave all possible flags in my bundle.yamls in case i need so set one. If it's empty it's faster to distinguish as not set. I'll remove it later.

PR #3186, opened 2026-03-20
Preserve view title position and extent when copying sheets

tay0thman:
Added functions to capture and apply viewport properties, including label offsets and bounding box corrections for viewports when copying sheets to open documents.

# Name of your PR
Preserve view title position and extent when copying sheets

## Description

When copying sheets to other open documents, the view title (label) for each viewport was placed at Revit's default offset instead of matching the source sheet. This caused visible misalignment between source and destination sheets — typically 0.3" to 2.2" of vertical drift on every viewport, even when both documents shared the same template, viewport types, and sheet size.

---

## Checklist

  ```bash
  ```

---

Problem
The original copy_sheet_viewports function placed viewports using Viewport.Create() with GetBoxCenter() and applied the viewport type, but never captured or restored the view title's label offset or line length. Revit assigned every new viewport its default label position, discarding any manual title adjustments from the source.
Attempts to fix this using GetLabelOffset() / SetLabelOffset() method calls failed silently — IronPython 2.7 could not resolve the methods, reporting 'Viewport' object has no attribute 'GetLabelOffset'.
Root Cause
LabelOffset and LabelLineLength on Autodesk.Revit.DB.Viewport are .NET properties (backed by get_/set_ accessors), not standalone methods. IronPython 2.7 exposes .NET properties as Python attributes — method-style calls like vport.GetLabelOffset() fail, while property-style access like vport.LabelOffset works correctly.
Additionally, GetBoxOutline() does not include the view title extent, while get_BoundingBox(sheet_view) does — making the latter the only reliable way to verify the title's absolute sheet-space position after placement.
Modified copy_sheet_viewports now follows a 5-step pipeline per viewport:

Capture source viewport data (center, label offset, line length, sheet bbox)
Viewport.Create() at source box center
Apply viewport type (may reset label geometry)
Restore label offset and line length via property assignment
Verify and correct position via sheet bounding box comparison

### DIFF
--- a/extensions/pyRevitTools.extension/pyRevit.tab/Drawing Set.panel/Sheets.pulldown/Copy Sheets to Open Documents.pushbutton/script.py
+++ b/extensions/pyRevitTools.extension/pyRevit.tab/Drawing Set.panel/Sheets.pulldown/Copy Sheets to Open Documents.pushbutton/script.py
@@ -408,19 +408,23 @@ def get_source_vport_data(doc, vport, sheet_view):
     }
     try:
         data["label_offset"] = vport.LabelOffset
-    except Exception:
-        pass
+    except AttributeError:
+        logger.debug("LabelOffset not available (requires Revit 2022+)")
+    except Exception as e:
+        logger.debug("Could not read LabelOffset: {}".format(e))
     try:
         data["label_line_length"] = vport.LabelLineLength
-    except Exception:
-        pass
+    except AttributeError:
+        logger.debug("LabelLineLength not available (requires Revit 2022+)")
+    except Exception as e:
+        logger.debug("Could not read LabelLineLength: {}".format(e))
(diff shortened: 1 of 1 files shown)


tay0thman:
@devloai , please review the logic update, tested locally and works great

tay0thman:
@devloai Please confirm your fixes and the latest commits are good to go and we are ready to merge.

jmcouffin:
@tay0thman did you test it after the latest fixes?

tay0thman:
@jmcouffin yes, I test it and we deployed it officewide. All good.
That was a complaint from my coworkers.

jmcouffin:
> @jmcouffin yes, I test it and we deployed it officewide. All good. That was a complaint from my coworkers.

Thanks mate, that really makes my life easier.

PR #3187, opened 2026-03-21
Update get_name() in query.py with worksets support

Denver-22:
A universal method, including support for obtaining a name from worksets.


jmcouffin:
added new unittest for the method since this is a fundamental one
image

PR #3299, opened 2026-04-12
Parity fixes for the new C# loader, plus Output Window robustness

romangolev:
PR Description

  ## Summary

  A grab-bag of fixes and improvements for the new C# session loader
  (`pyRevitAssemblyBuilder` + related runtime pieces). The common thread is
  bringing the C# loader into closer parity with the legacy Python loader, and
  fixing a few rough edges that only show up in real multi-extension setups.

  None of these changes touch the legacy Python loader path — everything is
  gated behind `new_loader = true`. Users on the old loader are unaffected.

  ## What changed and why

  ### 1. Extension authorization parity (`authusers` / `authgroups`)

  The legacy Python loader honours `authusers` / `authgroups` entries in
  `extension.json`: an extension is only loaded for users whose Windows username
  matches one of the `authusers`, or who belong to one of the `authgroups`
  (resolved via SID). The new C# loader didn't implement this at all, so
  authorization-restricted extensions were visible to everyone when
  `new_loader = true`.

  This PR:

  - Adds `AuthorizedUsers` and `AuthorizedGroups` to `ParsedExtension`.
  - Parses `authusers` / `authgroups` arrays from `extension.json` in the
    extension parser (matching how Python reads them).
  - Implements a `user_has_access` check in `ExtensionManagerService` that
    filters extensions by the current username and by group SID membership,
    so the behaviour matches the legacy loader exactly.
  - Pre-materialises library-extension lib paths during the same pass, which
    also shaves some load time.

  ### 2. Keep the session Output Window alive during startup scripts

  On a fresh session load, any extension with a startup script may open its own
  `ScriptConsole`. When the user has `close_other_outputs` enabled, each of
  those new consoles runs `ApplyCloseOthersConfig` on `Window_Loaded`, which
  closes *every other* open output window — including the session loader's own
  output window.

  That's fatal, because:

  - `ScriptConsole.Window_Closed` sets `ClosedByUser = true` regardless of who
    triggered the close (the name is misleading; it's really "closed by
    anyone").
  - From that point on, `ScriptIO.Write` short-circuits at
    `if (output.ClosedByUser) { return; }` and silently drops every write.
  - Everything logged after that — including e.g. a `StackBuilder` warning
    about a duplicate stack name — lands only in the log file, never in the
    Output Window.

  The fix introduces an `IsSessionOutput` flag on `ScriptConsole`. The session
  loader marks its own window with it, and `CloseActiveOutputWindows` skips
  any window that has it set, in both the exclude-this-window branch and the
  close-all branch. Startup scripts' `close_others` behaviour is preserved for
  regular script windows.

  ### 3. StackBuilder robustness + unit tests

  Improvements to how the stack builder assembles items and handles edge
  cases (fewer-than-2 items, existing stack items, duplicate names). New
  `StackBuilderTests` covers the paths, with a shared `MockLogger` helper so
  other parser tests can reuse it.

  ### 4. Build-system tweaks

  The Roslyn `Microsoft.CodeAnalysis*.dll` set was being copied into many
  output directories during build, slowing incremental builds and bloating
  the bin tree. `Directory.Build.targets` now skips redundant copies, and a
  new `DeployCodeAnalysisDlls` target handles the one location that still
  needs them. Net result: faster `pipenv run pyrevit build products` and
  smaller diffs on DLL commits.

  ### 5. Small related touch-ups

  - `AssemblyBuilderService` and `CommandTypeGenerator` — minor adjustments
    to support the above flows.
  - `SessionManagerService` — wiring for the parity/filtering work.
  - `AGENTS.md` — project-wide instruction for the opencode agent.

  ## Files of interest

  - `dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs`
  - `dev/pyRevitLoader/pyRevitExtensionParser/ParsedExtension.cs`,
    `dev/pyRevitLoader/pyRevitExtensionParser/ExtensionParser.cs`
  - `dev/pyRevitLabs.PyRevit.Runtime/ScriptConsole.cs`,
    `dev/pyRevitLabs.PyRevit.Runtime/ScriptConsoleManager.cs`
  - `pyrevitlib/pyrevit/loader/sessionmgr.py`
  - `dev/pyRevitLoader/pyRevitAssemblyBuilder/UIManager/Builders/StackBuilder.cs`
  - `dev/pyRevitLoader/pyRevitExtensionParserTester/StackBuilderTests.cs`,
    `MockLogger.cs`
  - `dev/pyRevitLoader/Directory.Build.targets`

  ## Notes for reviewers

  - All runtime changes are behind the `new_loader` flag. A legacy-loader
    reload should be byte-identical in behaviour.
  - The `IsSessionOutput` flag is additive — only the session loader sets it,
    so existing scripts that rely on `output.close_others()` see no change.
  - The `ClosedByUser` field name is retained for back-compatability


dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -45,34 +49,115 @@ public void ClearParserCaches()
...
+            if (ext.Config?.Disabled == true)
+            {
+                _logger?.Debug($"Extension '{ext.Name}' is disabled in config");
+                return false;
+            }
+
+            // Check authorized users list
+            if (ext.AuthorizedUsers != null && ext.AuthorizedUsers.Count > 0)
+            {
+                var currentUser = Environment.UserName;
```
tay0thman: may need to either pass the sanitized username as a parameter, or replicate the sanitization logic against Environment.UserName, The C# loader doesn't have easy access to the Revit Application object from ExtensionManagerService

dev/pyRevitLoader/pyRevitAssemblyBuilder/UIManager/SessionManagerService.cs
```diff
@@ -391,7 +410,8 @@ private void ExecuteExtensionStartupScript(ParsedExtension extension, List<Parse
         /// Builds the search paths for a startup script, including extension lib folders and pyRevit core paths.
         /// </summary>
         /// <param name="extension">The extension for which to build search paths.</param>
-        /// <param name="libraryExtensions">List of library extensions to include.</param>
+        /// <param name="libraryExtensions">List of library extensions to include (unused - kept for compatibility).
```
tay0thman: Since BuildSearchPaths is private, there's no external compatibility concern — the parameter could just be removed along with the argument at the call site

tay0thman:
I'm terrible at reviewing code, so once the authusers / authgroups precedence is fixed as well as fixing user name source with sanitation logic, I will clone the repo and stress test it. Thank you @romangolev

romangolev:
@tay0thman thanks, you can use this extension for testing. Add your pc name `$env:USERNAME` to the list of "authusers" and this will enable the extensions. 
[pyRevitTestExtensions.zip](https://github.com/user-attachments/files/26671346/pyRevitTestExtensions.zip)

Now thinking of it, probably it would make sense to notify to the output window if user does not have an access to the extension 🤔

Wurschdhaud:
As this PR is for feature parity: #3300 would fit great in as well, while you're on it.

romangolev:
@Wurschdhaud, that's a good one, thanks for pointing at it. I'll be on it right away

jmcouffin:
> As this PR is for feature parity: #3300 would fit great in as well, while you're on it.

Generally speaking Micro/scoped/focused PR are easier to review. And preferred...

romangolev:
> > As this PR is for feature parity: #3300 would fit great in as well, while you're on it.
> 
> 
> 
> Generally speaking Micro/scoped/focused PR are easier to review. And preferred...

In this case, there'll be another PR from my side 😄

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
romangolev: That a dealbreaker actually. I got it fixed now, passed Revit UI and now Revit username is being normalized and checked against the list from settings. Should be alright now.

dev/pyRevitLoader/pyRevitAssemblyBuilder/UIManager/SessionManagerService.cs
romangolev: Removed it

romangolev:
> I'm terrible at reviewing code, so once the authusers / authgroups precedence is fixed as well as fixing user name source with sanitation logic, I will clone the repo and stress test it. Thank you @romangolev

Fixed a couple of thing here and there. Ready for testing when you're ready, thanks!

dev/pyRevitLoader/pyRevitAssemblyBuilder/AssemblyMaker/CommandTypeGenerator.cs
```diff
@@ -72,21 +95,30 @@ public string GenerateExtensionCode(ParsedExtension extension, string revitVersi
                 .Where(d => !string.IsNullOrEmpty(d))
                 .ToList();
 
+            var totalCommands = extension.CollectCommandComponents().Count();
+
             foreach (var cmd in extension.CollectCommandComponents())
```
sanzoghenzo: Is the CollectCommandComponents an expensive operation? If so, it is better to use a local variable to compute it only once

dev/pyRevitLoader/pyRevitAssemblyBuilder/AssemblyMaker/CommandTypeGenerator.cs
```diff
@@ -72,21 +95,30 @@ public string GenerateExtensionCode(ParsedExtension extension, string revitVersi
...
+                    _logger.Error($"Skipped duplicate command: '{safeClassName}'. " +
+                                  $"Script: {cmd.ScriptPath ?? "(none)"}. " +
+                                  $"UniqueId: {cmd.UniqueId}. " +
+                                  $"Two bundle directories produced the same UniqueId. " +
+                                  $"Rename one directory to fix this.");
+
+                    skippedCommands.Add(new SkippedCommandInfo
+                    {
+                        SafeClassName = safeClassName,
+                        ScriptPath = cmd.ScriptPath,
```
sanzoghenzo: Doe this need a null check like done in the logger? The field is declared as non nullable if I read it right

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -11,11 +14,27 @@ namespace pyRevitAssemblyBuilder.SessionManager
     public class ExtensionManagerService : IExtensionManagerService
     {
         private readonly int _revitYear;
+        private readonly UIApplication? _uiApplication;
+        private readonly ILogger? _logger;
         private List<ParsedExtension>? _cachedExtensions;
+        private readonly HashSet<string> _authLogdedExtensions = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
sanzoghenzo: What does "Logded" means? Maybe "_authorizedExtensions" sounds better?

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -45,34 +65,131 @@ public void ClearParserCaches()
...
 
         /// <summary>
-        /// Gets all installed extensions that are not disabled.
+        /// Checks if the current user has access to the extension based on:
+        /// 1. Whether the extension is disabled in config
+        /// 2. Whether authorized users are defined and current user is in the list
+        /// 3. Whether authorized groups are defined and user is member of at least one
+        /// 
+        /// This replicates the logic in Python's extpkgs.ExtensionPackage.is_enabled + user_has_access
+        /// matching pyrevitlib/pyrevit/extensions/extensionmgr.py:_is_extension_enabled()
```
sanzoghenzo: No need to document this, IMHO. The whole goal of the new loader is to match (and surpass) the python loader in featuers, isn't it?

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -45,34 +65,131 @@ public void ClearParserCaches()
...
+                || (ext.AuthorizedGroups != null && ext.AuthorizedGroups.Count > 0);
+
+            if (!hasAuthRestrictions)
+                return true;
+
+            if (_authLogdedExtensions.Contains(ext.Name))
+                return true;
+
+            // Check authorized users list
+            if (ext.AuthorizedUsers != null && ext.AuthorizedUsers.Count > 0)
```
sanzoghenzo: This is used twice in few lines, it would be better to introduce a local or a property

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -45,34 +65,131 @@ public void ClearParserCaches()
...
+                    _authLogdedExtensions.Add(ext.Name);
+                    return false;
+                }
+                _logger?.Info($"User '{currentUser}' is authorized for extension '{ext.Name}'");
+                _authLogdedExtensions.Add(ext.Name);
+                return true;
+            }
+
+            // Check authorized groups list (groups are stored as SID strings) - only if no AuthorizedUsers restriction
+            if (ext.AuthorizedGroups != null && ext.AuthorizedGroups.Count > 0)
```
sanzoghenzo: Same here, use a local or property to make it clearer and reduce repetitions

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -45,34 +65,131 @@ public void ClearParserCaches()
...
+            {
+                bool inAnyGroup = false;
+                foreach (var groupSid in ext.AuthorizedGroups)
+                {
+                    if (UserIsInSecurityGroup(groupSid))
+                    {
+                        inAnyGroup = true;
+                        break;
+                    }
+                }
```
sanzoghenzo: This could be Simplified with Linq's Any

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -45,34 +65,131 @@ public void ClearParserCaches()
...
+                if (!inAnyGroup)
+                {
+                    _logger?.Warning($"Extension '{ext.Name}' is NOT available for current user (not in AuthorizedGroups)");
+                    _authLogdedExtensions.Add(ext.Name);
```
sanzoghenzo: I believe this is an error, it adds the extension eve if not authorized

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -45,34 +65,131 @@ public void ClearParserCaches()
...
+                _logger?.Info($"User is authorized for extension '{ext.Name}' (AuthorizedGroups match)");
+                _authLogdedExtensions.Add(ext.Name);
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Checks if the current Windows user is a member of the specified security group.
+        /// Uses the same implementation as pyRevitLabs.Common.Security.UserAuth.UserIsInSecurityGroup.
```
sanzoghenzo: I surely am missing something, but why reimplement the code instead of calling it from that assembly?

sanzoghenzo (changes requested):
I think I found a bug...
The rest are mostly my preferences, and a naming issue

dev/pyRevitLoader/pyRevitAssemblyBuilder/AssemblyMaker/CommandTypeGenerator.cs
romangolev: This function uses cached data, so I wouldn't consider it as superexpensive. Regardless, that's a good idea to store it's result as local variable ^

dev/pyRevitLoader/pyRevitAssemblyBuilder/AssemblyMaker/CommandTypeGenerator.cs
romangolev: Made the struct fields nullable

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
romangolev: That's true. Fully agreed, that's redundunt

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
romangolev: That's obviusly a typo, meant to be Logged. The authorized sounds simpler and better

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
romangolev: Good one, reused the on from pyrevitlabs

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
romangolev: resolved

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -41,38 +63,143 @@ public void ClearCache()
...
+                _logger?.Warning($"Extension '{ext.Name}' is NOT available for current user (not in AuthorizedGroups)");
+                return false;
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Checks if the current Windows user is a member of the specified security group.
+        /// Uses pyRevitLabs.Common.Security.UserAuth.UserIsInSecurityGroup with local guards.
```
sanzoghenzo: Not a fan of describing implementation details in documentation, can you please remove this?

dev/pyRevitLoader/pyRevitAssemblyBuilder/AssemblyMaker/CommandTypeGenerator.cs
```diff
@@ -72,21 +95,31 @@ public string GenerateExtensionCode(ParsedExtension extension, string revitVersi
...
-                    // which script was dropped (the .cs file is saved alongside the DLL).
-                    sb.AppendLine($"// WARNING [#3140]: Skipped duplicate class '{safeClassName}'");
-                    sb.AppendLine($"//   Script: {cmd.ScriptPath ?? "(no script)"}");
-                    sb.AppendLine($"//   UniqueId: {cmd.UniqueId}");
-                    sb.AppendLine($"//   Two bundle directories produced the same UniqueId.");
-                    sb.AppendLine($"//   Rename one directory to fix this.");
-                    sb.AppendLine();
+                    // Log an error so user sees it, and track for summary.
+                    _logger.Error($"Skipped duplicate command: '{safeClassName}'. " +
+                                  $"Script: {cmd.ScriptPath }. " +
```
sanzoghenzo: Here is preferable to keep the null check to display a proper message ("No script" or similar) instead of "null" (see also the comment above for the handling of the other fields)

dev/pyRevitLoader/pyRevitAssemblyBuilder/AssemblyMaker/CommandTypeGenerator.cs
```diff
@@ -17,10 +17,30 @@ namespace pyRevitAssemblyBuilder.AssemblyMaker
...
     public class RoslynCommandTypeGenerator
     {
+        private readonly SessionManager.ILogger _logger;
+
+        private struct SkippedCommandInfo
+        {
+            public string? SafeClassName;
+            public string? ScriptPath;
+            public string? UniqueId;
+            public string? BundleDirectory;
```
sanzoghenzo: I didn't check the command class from where these info comes from (I'm on mobile, and it's hard to navigate this codebase), but are they all nullable?
From the previous code, it seems that only the path can be null... Sorry if I wasn't clear earlier and disregard this message if they are indeed all nullable. But if this is the case, the log message below should handle null values

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -41,38 +63,143 @@ public void ClearCache()
...
+
+            return isAllowed;
+        }
+
+        private bool EvaluateAuthorization(
+            ParsedExtension ext,
+            List<string>? authorizedUsers,
+            List<string>? authorizedGroups,
+            bool hasAuthorizedUsers,
+            bool hasAuthorizedGroups)
```
sanzoghenzo: Hmm, now that you extracted the method, it makes little sense to use and pass the boolean variables since they are not independent and can (should) be deducted from the lists.
This signature allows to pass a null ot empty string but a true value for the check and vice versa,  not a great design!

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -41,38 +63,143 @@ public void ClearCache()
...
+        /// 3. Whether authorized groups are defined and user is member of at least one
+        /// </summary>
+        private bool IsExtensionAllowed(ParsedExtension ext)
+        {
+            var authorizedUsers = ext.AuthorizedUsers;
+            var authorizedGroups = ext.AuthorizedGroups;
+            var hasAuthorizedUsers = authorizedUsers is { Count: > 0 };
+            var hasAuthorizedGroups = authorizedGroups is { Count: > 0 };
+
+            if (ext.Config?.Disabled == true)
```
sanzoghenzo, line 79:
```suggestion
            if (ext.Config?.Disabled)
```

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -41,38 +63,143 @@ public void ClearCache()
...
+        /// 1. Whether the extension is disabled in config
+        /// 2. Whether authorized users are defined and current user is in the list
+        /// 3. Whether authorized groups are defined and user is member of at least one
+        /// </summary>
+        private bool IsExtensionAllowed(ParsedExtension ext)
+        {
+            var authorizedUsers = ext.AuthorizedUsers;
+            var authorizedGroups = ext.AuthorizedGroups;
+            var hasAuthorizedUsers = authorizedUsers is { Count: > 0 };
+            var hasAuthorizedGroups = authorizedGroups is { Count: > 0 };
```
sanzoghenzo: These can be moved below the first conditional to save a bit of processing if the extension is disabled

dev/pyRevitLoader/pyRevitAssemblyBuilder/SessionManager/ExtensionManagerService.cs
```diff
@@ -41,38 +63,143 @@ public void ClearCache()
...
+
+            if (ext.Config?.Disabled == true)
+            {
+                _logger?.Debug($"Extension '{ext.Name}' is disabled in config");
+                return false;
+            }
+
+            if (!hasAuthorizedUsers && !hasAuthorizedGroups)
+                return true;
```
sanzoghenzo: Thinking out loud here: what if we move this inside the EvaluateAuthorization method? Sure, there will be two more conditionals to check each time (when the extension is not cached), but it will simplify the lists checks (see comment below)

sanzoghenzo (changes requested):
Somehow I led you to worsen things (from my point of view). Sorry about that...

jmcouffin:
@romangolev let me know when ready for testing

romangolev:
> Somehow I led you to worsen things (from my point of view). Sorry about that...

No worries, also my bad. Got the code cleaned up a little once more. Hope it's good enough now to hit the develop. Regardless, we fixed a bug and optimized one function, so all good 🙌

romangolev:
> @romangolev let me know when ready for testing

Ready when you are 😄 
Mostly some cosmetics that I've been editing recently.

dev/pyRevitLoader/pyRevitAssemblyBuilder/AssemblyMaker/CommandTypeGenerator.cs
```diff
@@ -72,21 +95,31 @@ public string GenerateExtensionCode(ParsedExtension extension, string revitVersi
...
+                    _logger.Error($"Skipped duplicate command: '{safeClassName}'. " +
+                                  $"Script: {cmd.ScriptPath ?? "<none>"}. " +
+                                  $"UniqueId: {cmd.UniqueId}. " +
+                                  $"Two bundle directories produced the same UniqueId. " +
+                                  $"Rename one directory to fix this.");
+
+                    skippedCommands.Add(new SkippedCommandInfo
+                    {
+                        SafeClassName = safeClassName,
+                        ScriptPath = cmd.ScriptPath,
```
sanzoghenzo: Sorry for being pedantic, but now we're back to square one... here cmd.ScriptPath can be null, but the SkippedCommandIno.ScriptPath is not nullable, so we need something like `cmd.ScriptPath ?? "<none>"` here too

sanzoghenzo (commented):
Maybe it's a non-problem, but this is my last complaint 😉

dev/pyRevitLoader/pyRevitAssemblyBuilder/AssemblyMaker/CommandTypeGenerator.cs
romangolev: removed them from struct entirely

sanzoghenzo (approved):

jmcouffin (approved):
Massive!
Thanks
Remember, smaller PR is better @romangolev

romangolev:
> Massive!
> Thanks
> Remember, smaller PR is better @romangolev 

This one could have been splitted at least in two! Totally fair comment 👍

jmcouffin:

PR #3313, opened 2026-04-17
Add Seed43 extension to extensions.json

Seed-43:
# Name of your PR

Add Seed43 extension to extensions.json

## Description

Adds the Seed43 PyRevit extension to the community extensions registry.

Seed43 is an extension for Revit 2025/2026 with an in-app manager to install and update 
Seed43 tools directly from GitHub.

That includes PyTransmit (transmit Revit files with automatic sheet and link management)
and 3D Tools (3D workflow tools for Revit).

---

## Checklist

  ```bash
  ```

---

## Related Issues

- Resolves #[issue number]

---

## Additional Notes

Extension repository: https://github.com/Seed-43/Seed43
Website: https://seed43.org

---


jmcouffin:
@Seed-43 Can you clarify a bit here?
The repo points to an about tool and a reload, right?
pyrevitTransmit, for example, does not exist anywhere in your github org?

your repo structure is also misleading with a .extension folder and a .tab at the same level.
external extension should always start with a .tab folder only.

Seed-43:
@jmcouffin thanks for your reply.

To clarify the repo does points to an about tool and a reload. This is a tool to install the Seed43 tools the users would like to use, Instead of me pushing everything I developed onto the user.

I Have removed the .extension folder and keept the .tab
This was only like this because I didn't realise I only needed the .tab and I needed to remap the About installer to the new .tab folder at the office. Otherwise the update logic would not work.

So to recap I removed .extensions
And the about is an installer that copies user preferred tools for Seed43. It also has intelligence to notify users of an update and will update when the user clicks to accept.

Regards.

Seed-43:
I have cleaned up the folder and removed the reload tool as this was only for testing purposes.

Should be only one tool, it is the installer for the Seed43 tools, users can install other tools from this tool.

jmcouffin:
I am not exactly comfortable endorsing na extension that does not expose its tools. It feels like na opened door for 🕳️

Seed-43:
Ok, I understand, I will update the tools to allow users to unload tools that they don't need and add the tools I have to the repo so it's opt out instead of opt in. I will close this commit and do a new one when I have had time to update everything.

Issue #2823, opened 2025-09-28
Use six in pyrevitlib


sanzoghenzo:
Following #2811 

we could use six to ensure compatibility with both python 2 and 3, instead of manually patching pyrevitlib's code.

As of now, the problem to introduce this is that site-packages, that is where six is placed, is not loaded by PyRevitLoaderApplication (as opposed to PyRevitRunner that loads both pyrevitlib and site-packages before running the user script).

We have two possible options:

1. add site-packages to the syspaths in PyRevitLoaderApplication
2. copy over the six module inside pyrevitlib. From the documentation:
    >  six consists of only one Python file, so it is painless to copy into a project

My thoughts: since #2647 is on the way, both options should be temporary, so option 1 would be better (no need to copy things around, and no need to change imports afterwards). But I don't know if this would introduce unwanted side effects.

PR #2647, opened 2025-04-14
Rewrite Loader into C# 

romangolev:
# Rewrite Loader into C# 

## Description

Implementing an AssemblyBuilder, UIBuilder and SessionManager refactored in C# as a substitute for the original Loader which is written in Python. 
This feature aims to increase the load speeds on startup and reload events.

---

## Checklist

Before submitting your pull request, ensure the following requirements are met:

- [ ] Code follows the [PEP 8](https://peps.python.org/pep-0008/) style guide. -- Not applicable
- [ ] Code has been formatted with [Black](https://github.com/psf/black) using the command: -- Not applicable
  ```bash
  pipenv run black {source_file_or_directory}
  ```  
- [ ] Changes are tested and verified to work as expected. - No, just a draft for now

---

## Related Issues

If applicable, link the issues resolved by this pull request:

- N/A

---

## Additional Notes

THIS IS A DRAFT
Entry point is at PyRevitLoaderApplication which now have 2 modules for loading pyRevit sessions: 
- legacy: ExecuteStartUpPython
- refactored: ExecuteStartUpCsharp

Adds new project pyRevitAssemblyBuilder which is set as a dependency project for pyRevitLoader and pyRevitRunner. 
Whenever one of them built, dependency is built and loaded automatically. 

pyRevitAssemblyBuilder is using DI (Dependency Injection) and have a set of services being instantiated on Revit startup as singletons.

On the moment of opening this draft PR, the module is able to build dummy dlls and simplified UI for them (without images and tooltips). Yet does not support yaml bundle files, does not build lib files. 

## Reviewers
@jmcouffin @sanzoghenzo @dosymep what do you think? Is it a way to go? 

---

Thank you for contributing to pyRevit! 🎉

PR #3319, opened 2026-04-19
Enable six at Loader Startup by Adding site-packages to sys.path 

romangolev:
Add `site-packages` to startup loader `sys.path`

This is more of a discussion to the topic that had already been opened for a while, 
mostly regarding 
- #2823

I believe that [pyRevitLoader.py](https://github.com/pyrevitlabs/pyRevit/compare/feat/six?expand=1#diff-78826b1d1812b8f6fe166a48f7986dc8309457cfe96d34205e859992ec9ed737) would be the place where we need to import six in order for it to be available at the startup. 

Here are some of the agentic summary in order for everyone to get onboard with the problem: 

## Summary

Appends `site-packages` to `sys.path` in `pyRevitLoader.py` (both `bin/netfx/` and `bin/netcore/` variants), so packages vendored under `site-packages/` — notably `six` — become importable from `pyrevitlib` code that runs at Revit session-load time.

This is a **prerequisite enabler** for the follow-up refactor of `pyrevitlib/pyrevit/compat.py` to use `six` (issue #2811) instead of hand-maintained `if PY2 / elif PY3` branches for `winreg`, `urllib`, `unicode`, `Iterable`/`Callable`, etc.

Related: #2811 (use `six` for PY2/PY3 unification).

## Why this is needed

`compat.py` is imported by `pyrevit/__init__.py` at session load — i.e. while `pyRevitLoader.py` is initialising the startup IronPython engine, **before** any user script (and therefore any executor) has run.

Audit of how `site-packages` reaches each engine:

| Stage                                  | Adds `pyrevitlib` | Adds `site-packages` |
|----------------------------------------|-------------------|----------------------|
| Startup (`pyRevitLoader.py`)           | yes               | **no** ← gap         |
| User scripts (`PyRevitRunnerCommand.cs:43-44`) | yes       | yes                  |
| Hooks (`HookManager.cs:179-188`)       | yes               | yes                  |
| Smart/ComboBox (`SmartButtonExecutor.cs:179-185`, `ComboBoxExecutor.cs:278-283`) | yes | yes |
| Generated commands (`CommandTypeGenerator.cs:95-130`) | yes | yes               |

Every executor already exposes `site-packages`. Only the startup loader did not, which is why `import six` at the top of `compat.py` failed at session-load time even though `site-packages/six.py` is present.

## Change

Two byte-identical files, one new line each, immediately after the existing `pyrevitlib` append:

```python
# bin/netfx/engines/pyRevitLoader.py
# bin/netcore/engines/pyRevitLoader.py
repo_path = op.dirname(op.dirname(op.dirname(op.dirname(__file__))))
sys.path.append(op.join(repo_path, 'pyrevitlib'))
sys.path.append(op.join(repo_path, 'site-packages'))   # NEW
```

`pyrevitlib` remains earlier on `sys.path` than `site-packages`, so pyrevit's own modules continue to take precedence (no shadowing risk for pyrevit names).

## Why not a different approach

- **Vendor `six.py` into `pyrevitlib/`** — creates a maintained out-of-tree copy that has to be undone after #2647, and breaks the convention that `pyrevitlib/` only holds pyrevit-authored code.
- **Append inside `pyrevit/__init__.py`** — `__init__.py` imports `from pyrevit.compat import PY3` at line 22, *before* its own `sys.path.append(BIN_DIR)` calls at lines 97-99. The append would happen too late for `compat.py` itself to use `six`.
- **Move the path injection into the C# loader** — invasive, requires a rebuild to test, and offers no functional benefit over the Python-side append (the C# add-in does not currently set `sys.path`; the Python loader script does).

## Why the runner-side `site-packages` adds are kept

Each script engine has its **own isolated `sys.path`**, not inherited from the loader engine:

- `IronPythonEngine.SetupSearchPaths` calls `Engine.SetSearchPaths(...)` which **replaces** the engine path list per script run (`dev/pyRevitLabs/pyRevitLabs.PyRevit.Runtime/IronPythonEngine.cs:284-286`).
- Engines are cached per-extension, not globally (`ScriptEngineManager.cs:6-22`). The startup engine instance is not the script-execution engine instance.
- `CPythonEngine` is a separate interpreter entirely; the loader's IronPython `sys.path` is irrelevant to it.

So the loader-side append and the runner-side adds are **not duplicates** — they feed two different engine populations. Removing the runner-side calls would break `import six`/`requests`/`sqlalchemy`/etc. for every user script. Out of scope for this PR.

## Risks

- **Latent surface widens**: every package under `site-packages` (`requests`, `urllib3`, `sqlalchemy`, `pytz`, …) is now reachable from session-load context. None are imported at startup today, so actual exposure is zero. A short comment on the new line warns future contributors against doing `import requests` (or similar heavy package) at module top in `pyrevitlib`, which would push DLL loads / network probes into Revit's startup window.
- **No assembly-load-order impact**: the change is in the Python script that already runs under IronPython, not in the C# add-in code object that `PyRevitLoaderApplication.cs` warns about.
- **Sync between netfx/netcore copies**: pre-existing tech debt; both files must be edited together.

## Follow-up (separate PR)

Refactor `pyrevitlib/pyrevit/compat.py` to use `six` for the PY2/PY3 branches now that `import six` is reachable at session-load time.

### DIFF
--- a/bin/netcore/engines/pyRevitLoader.py
+++ b/bin/netcore/engines/pyRevitLoader.py
@@ -25,6 +25,7 @@ import os.path as op
 # add the library location to the system search paths
 repo_path = op.dirname(op.dirname(op.dirname(op.dirname(__file__))))
 sys.path.append(op.join(repo_path, 'pyrevitlib'))
+sys.path.append(op.join(repo_path, 'site-packages'))
 
 # now pyrevit can be imported
 from pyrevit.loader import sessionmgr
--- a/bin/netfx/engines/pyRevitLoader.py
+++ b/bin/netfx/engines/pyRevitLoader.py
@@ -25,6 +25,7 @@ import os.path as op
 # add the library location to the system search paths
 repo_path = op.dirname(op.dirname(op.dirname(op.dirname(__file__))))
 sys.path.append(op.join(repo_path, 'pyrevitlib'))
+sys.path.append(op.join(repo_path, 'site-packages'))
 
 # now pyrevit can be imported
 from pyrevit.loader import sessionmgr


romangolev:
@Wurschdhaud it does affect places you've been working on recently

sanzoghenzo (approved):
While I dream the day the python loader is not needed anymore, I agree that this is the most effective way to deal with this issue!

jmcouffin:

Issue #3305, opened 2026-04-16
[Bug]: Pulldowns are not updated on reload


ING-MOP:
### ✈ Pre-Flight checks

- [x] I **don't have SentinelOne** antivirus installed (see above for the solution)
- [x] I have **searched in the issues** (open and closed) but couldn't find a similar issue
- [x] I have **searched in the [pyRevit Forum](https://discourse.pyrevitlabs.io)** for similar issues
- [ ] I already **followed the [installation troubleshooting guide](https://pyrevitlabs.notion.site/Installation-issues-f6495adb56254e9e8021f8a68e3b2ab7) thoroughly**
- [x] I am using the **latest pyRevit Version**

### 🐞 Describe the bug

When i click Reload pyRevit it would normally update all addins.
The new issue after V6 is:
If i add a new button to an existing pulldown, and i click reload pyRevit, it will not show the newly added button in the pulldown.
If i move the button from the pulldown to the panel instead of a pulldown, it will show when i click reload.

### ⌨ Error/Debug Message

```shell
No errors messages.
```

### ♻️ To Reproduce

1. Make sure you have a pulldown with some buttons
2. Open Revit
3. Copy one of the buttons and rename it to Test
4. Click reload pyRevit
5. Here it should add the newly created button, but it dosnt.

### 🖥️ Hardware and Software Setup (please complete the following information)

```shell
==> Registered Clones (full git repos)
==> Registered Clones (deployed from archive/image)
master | Branch: "master" | Version: "6.4.0.26100+0515" | Path: "C:\Users\mop\AppData\Roaming\pyRevit-Master"
==> Attachments
master | Product: "2026 First Customer Ship" | Engine: DEFAULT (2712) | Path: "C:\Users\mop\AppData\Roaming\pyRevit-Master"
master | Product: "2025 First Customer Ship" | Engine: DEFAULT (2712) | Path: "C:\Users\mop\AppData\Roaming\pyRevit-Master"
master | Product: "2024.3" | Engine: DEFAULT (2712) | Path: "C:\Users\mop\AppData\Roaming\pyRevit-Master"
master | Product: "2023.1.4" | Engine: DEFAULT (2712) | Path: "C:\Users\mop\AppData\Roaming\pyRevit-Master"
master | Product: "2022.1.7" | Engine: DEFAULT (2712) | Path: "C:\Users\mop\AppData\Roaming\pyRevit-Master"
[...]

Issue #3114, opened 2026-02-18
[Bug]: Dynamo Scripts needs to be set to "Automatic" in order to work


tay0thman:
### ✈ Pre-Flight checks

- [x] I **don't have SentinelOne** antivirus installed (see above for the solution)
- [x] I have **searched in the issues** (open and closed) but couldn't find a similar issue
- [x] I have **searched in the [pyRevit Forum](https://discourse.pyrevitlabs.io)** for similar issues
- [x] I already **followed the [installation troubleshooting guide](https://pyrevitlabs.notion.site/Installation-issues-f6495adb56254e9e8021f8a68e3b2ab7) thoroughly**
- [x] I am using the **latest pyRevit Version**

### 🐞 Describe the bug

The bug is related to : https://github.com/pyrevitlabs/pyRevit/issues/3076#issuecomment-3923629854

Dynamo scripts set to manual run won't be executed by pyRevit

### ⌨ Error/Debug Message

```shell
None
```

### ♻️ To Reproduce

1. create a dynamo script.
2. set the run method to "Manual"
3. save as script.py and place in a pushbutton
4. reload pyrevit
5. click to run the script
6. nothing happens

### ⏲️ Expected behavior

It is expected that the script will run as automatic always, even when the run method is set to manual

### 🖥️ Hardware and Software Setup (please complete the following information)

```shell
Revit 2026.4 , Latest Dynamo Core, Pyrevit 6.1
```

tay0thman:
@jmcouffin @romangolev  
I think it is related to the new loader.

jmcouffin:
https://github.com/pyrevitlabs/pyRevit/issues/3076#issuecomment-3925268411

PR #3328, opened 2026-04-23
fix(newloader): apply beta + version filter to all UI components

romangolev:
A fix for the following issue: 
- #3305

 ## Summary                                                                
  - Beta and `min/max_revit_version` filtering now applies to children of 
  pulldown, split, and stack components — previously only the top-level
  component was checked, so beta children stayed visible inside non-beta    
  parents.
  - Reload now creates newly-visible children when `LoadBeta` or version    
  constraints change between sessions, and deactivates children that became 
  hidden. `SplitButton.CurrentButton` is rebound when its previously-active 
  child is hidden, so the primary action no longer renders as nothing.      
  - Refactor: builders now share a per-build settings snapshot via a        
  `BuildContext` (single source of truth, no more independent
  `PyRevitConfig.Load()` calls); duplicate filter / short-circuit logic in  
  pulldown/split builders is collapsed into one helper; the deploy target   
  stops duplicating ~30 shared DLLs (Roslyn/VB, BCL, LibGit2Sharp,
  MadMilkman, OpenMcdf, `pyRevitLabs.Common`) into per-engine folders since 
  the loader already loads them from `bin/netcore/` (or `bin/netfx/`) at    
  startup.
  - Adds `ComponentSupportUtilsTests` (16 cases) covering beta/version      
  filtering, separator collapsing, and version-string normalization.


romangolev:
- fixes #3305

This bug should be covered as well by this PR. 
@jmcouffin sorry for not splitting them in chunks. They are highly interlaced together.

tay0thman:
@romangolev 

## Review notes — cross-version testing on NETFX + NETCORE

Tested with the committed DLLs on:

- **Revit 2024** (NETFX, net48)
- **Revit 2026** (NETCORE, .NET 8)
- **Revit 2027** (NETCORE — runs the net8.0-windows DLLs without issue on .NET 10)

Test fixture: a `BetaSplit.splitbutton` with three pushbutton children (`A_BetaFirst.pushbutton` with `is_beta: true`, `B_VisibleSecond.pushbutton`, `C_VisibleThird.pushbutton`) placed inside a non-beta panel alongside a non-beta sibling pushbutton, so the panel itself stays populated regardless of beta state.

### What works ✓

- **Beta filter cascades through nested UI components.** Beta children inside non-beta pulldowns now hide correctly. When all children of a pulldown are filtered out, the pulldown itself collapses; same for panels with no remaining visible content. The recursive walk handles arbitrary nesting depth.

- **Reload toggling is fully bidirectional for create/destroy.** Toggling `load_beta` between sessions and reloading correctly creates newly-visible children and deactivates newly-hidden ones, no Revit restart required. Both `is_beta: true` in `bundle.yaml` and module-level beta declarations behave identically (one-flag-many-sources working as designed).

- **Panel-level `is_beta` cascades correctly.** Caught one of my own panels with this property set during testing — when `load_beta=false`, the entire panel disappears as expected, including all its children. Useful sanity signal that the cascade also triggers at higher levels of the tree.

- **SplitButton CurrentButton rebinds on hide** (commit 462bfa5d3). When the first child is filtered out, `CurrentButton` correctly rebinds to the next visible child, and clicking the primary face dispatches to that child rather than the hidden one.

- **Build deduplication works.** No `Could not load file or assembly` errors in pyRevit logs. The shared DLLs (`LibGit2Sharp`, `pyRevitLabs.Common`, Roslyn/VB, MadMilkman, OpenMcdf, BCL, etc.) are correctly resolved from `bin/netcore/` and `bin/netfx/` with no per-engine duplication.

### !!!Asymmetric SplitButton rebind on the show direction

The `CurrentButton` rebind only fires on hide, not on show. Toggling `load_beta` back to `true` after the first child becomes visible again does not restore it as `CurrentButton`.


**Repro** (using the fixture above):

1. `load_beta = true`, reload → CurrentButton = `A_BetaFirst` ✓ image
2. `load_beta = false`, reload → CurrentButton rebinds to `B_VisibleSecond` ✓ *(this is the headline fix)*image
3. `load_beta = true`, reload → CurrentButton **remains** `B_VisibleSecond` ✗image


**** Expected** (legacy parity): CurrentButton restores to `A_BetaFirst` since A is visible again and precedes B in declaration order.**

The legacy Python loader rebuilt SplitButtons from scratch on every reload, so this asymmetry never existed there. The new incremental reload correctly handles "previously-active child became hidden" but doesn't re-evaluate when "a previously-hidden, higher-precedence child becomes visible." Looks like it needs a complementary check in the reload branch of `SplitButtonBuilder` — if a newly-visible child precedes `CurrentButton` in the declared child order, rebind to it.

Confirmed identical behavior on all three Revit versions tested, so it's not runtime-specific.

### Smoke regression — clean

Ribbon loads cleanly, About dialog shows correct version, normal pushbuttons execute, Shift+Click opens config scripts, Dynamo `.dyn` buttons execute (regression check vs #3114 fix), `pyrevit env` reports correctly, and no assembly resolution errors in logs related to the deduplicated DLLs.

### Verdict

The four claimed fixes all work as advertised. The show-direction rebind is the only outstanding behavior I'd want addressed before merge — it's more of a legacy parity regression than a new bug, but it's user-visible and likely a small change scoped to the same builder file.

romangolev:
@tay0thman good catch. I've pushed the update with a fix, take a look at it please 🙌

tay0thman:
> @tay0thman good catch. I've pushed the update with a fix, take a look at it please 🙌

Awesome! The Latest Commit Resolved all Issues , Tested Locally! Way to go @romangolev LGTM.

tay0thman (approved):
Tested Locally and all Issues Resolved

jmcouffin:

PR #3337, opened 2026-04-27
fix(config): prevent telemetry config escape-doubling

### LINKED ISSUE #3334: [Bug]: Telemetry config fields exhibit progressive escape-doubling, growing to 8MB+ each over reload cycles
### ✈ Pre-Flight checks

### 🐞 Describe the bug

Summary
Three fields in the [telemetry] section of pyRevit_config.ini accumulate escape characters on every save, doubling in size each cycle. After ~22 reload cycles they reach approximately 8 MB each, bloating the config file to ~25 MB. This dramatically slows PyRevitConfig.Load() and indirectly slows pyRevit reload to several minutes (vs ~40s for a clean config).

Image

Affected fields

telemetry_file_dir
telemetry_server_url
apptelemetry_server_url

Other telemetry fields in the same section (utc_timestamps, active, active_app, apptelemetry_event_flags, include_hooks) are not affected. Other string fields elsewhere in the config (extension paths, clone paths, tab colors, last window geometry) are also not affected — the bloat is specific to these three.

### ⌨ Error/Debug Message

```shell
After many cycles, the value looks like:
telemetry_server_url = "\"\\\"\\\\\\\"\\\\\\\\\\\\\\\"\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\"..."
```

### ♻️ To Reproduce

1. Open %APPDATA%\pyRevit\pyRevit_config.ini
2. Note the current value of telemetry_server_url (or set it to anything containing characters that need JSON-escaping, such as a URL with /)
3. Open Revit with pyRevit attached → reload pyRevit → close Revit
4. Inspect the value — additional escape sequences will have been added to the stored representation
5. Repeat steps 3–4 several times — the value grows exponentially

### ⏲️ Expected behavior

_No response_

```shell
pyRevit: 6.4.0.26098+2214 (current develop branch tip)
Revit: 2024.3.4 / 2026 FCS / 2027
OS: Windows 10 build 26200
CLI version: matches develop tip
```

### Additional context

The escape pattern doubles in depth on each save: \" → \\\" → \\\\\\\" → \\\\\\\\\\\\\\\" etc. This indicates the writer is re-escaping previously-escaped characters as if they were literal content of the value.
Impact
Even with telemetry fully disabled (active = false and active_app = false, which is the default), the bloated values still get parsed on every PyRevitConfig.Load(). With ~25 MB of escape-heavy text to process, this:

Adds 30–60s to first config read (cold cache)
Likely triggers pathological backtracking in whatever regex/parser handles these specific values
Cascades into reload performance — observed ~3 min reload vs ~40s baseline, where the only difference between baseline and bloated state was the size of these three fields

Reload time was fully restored to ~40s after blanking the three fields manually, confirming these values are the sole cause.
Workaround
Edit the config file directly while Revit is closed, set the three fields to empty strings:
initelemetry_file_dir = ""
telemetry_server_url = ""
apptelemetry_server_url = ""
Since telemetry is disabled by default, these values serve no functional purpose for most users. The bloat will resume on subsequent saves until the underlying serialization bug is fixed.
Suggested investigation area
The escape mismatch likely lives in the INI serializer used for these specific fields — possibly in pyRevitLabs.Common or wherever string fields with embedded special characters get persisted. The asymmetry is between:

Read path: does not unescape the previously-written escape characters (treats them as literal value content)
Write path: does JSON-encode the value before writing (adding another layer of escapes on top of what was read)

The result is that the value grows by one escape level per save cycle.
The fact that other string fields don't bloat suggests these three go through a different code path — perhaps a JSON-typed field that was unintentionally double-wrapped, or a string-typed field that was escape-encoded once at creation and the writer doesn't recognize the existing encoding.

tay0thman:
# fix(config): prevent telemetry config escape-doubling

## Description

Three telemetry config fields (telemetry_file_dir, telemetry_server_url, apptelemetry_server_url) accumulated JSON escape characters geometrically on every reload, growing each value to ~8 MB after ~22 cycles and slowing reload from ~40s to ~3 min.

Root cause: asymmetric encode/decode in
PyRevitConfigSectionParser. __setattr__ unconditionally json.dumps-wraps the value, while __getattr__'s outer fallthrough returns a fallback-chain-mutated string when all three json.loads attempts fail. The mutated value gets re-encoded on the next write, compounding on every reload via setup_telemetry() and upgrade_user_config().

Three changes:

1. coreutils/configparser.py
   - __setattr__ adds an idempotency check that skips the write when the encoded value matches what's already stored.
   - __getattr__ returns the original raw bytes (not the mutated fallback 'value') when all decode attempts fail.

2. versionmgr/upgrade.py
   - Adds heal_bloated_telemetry_fields() to detect and reset bloated fields on the next reload, with a backup of the original config saved as <config>.bloated.<timestamp>.bak.
   - Called from upgrade_user_config() before the format-upgrade loop so multi-megabyte values never round-trip the loop.

3. telemetry/__init__.py
   - Adds persist=True parameter to set_telemetry_file_dir, set_telemetry_server_url, set_apptelemetry_server_url.
   - setup_telemetry() passes persist=False, eliminating the redundant round-trip read-write cycle that previously fed the bug. Settings GUI save path keeps default persist=True.

image

Verified against a real 24 MB bloated config:
- Heal correctly reduced live config from 24,580 KB -> 4 KB on first launch, with backup preserved.
- Steady-state reload cycles produce 0-byte size delta (Fix A idempotency check working as designed).
- All three affected fields reset to "" post-heal.

---

## Checklist

---

## Related Issues

- Resolves #3334 

---

## Additional Notes


jmcouffin:

PR #3339, opened 2026-04-27
Improve recall — Match Properties recall window + select_parameters enhancements

### LINKED ISSUE #2840: Match Properties Reuse
## Is your feature request related to a problem? Please describe

I often have to copy engineer notations for beam Camber and Number of Studs. These numbers are often repeatable across a beam system. 

## Describe the solution you'd like

The "Match > Match Properties > Elements" tool works, but when I need to grab a different set of Camber and Number of Studs I need to re-select the properties from a long list. It would be nice if the tool remembered which properties you matched from the same element type such as structural columns or structural framing.

## Describe alternatives you've considered

Another option would be to have a CTRL + SHIFT (rather than just SHIFT) to reselect the same parameters that you want to copy.

## Additional context

Thank you for your time.

### LINKED ISSUE #2977: Maintain Matching Instance Parameters
Maintain the chosen parameters when repeating the command instead of repeatedly choosing the parameters.

Wurschdhaud:
## Description

- Recall window: Shift+Click on Match Properties now opens a modeless WPF window pre-filled with the last-matched parameters (all ticked), instead of immediately reapplying them. The window saves the current selection back to the memory file on close and restores its position between sessions.
- Reusable clipboard UI: The dockable clipboard panel and recall window share a single ClipboardContent(UserControl) backed by clipboard_content.xaml. The panel is now a thin Page wrapper; the recall window is a thin WPFWindow wrapper.
- forms.select_parameters – preselect + groups: The function gains a preselect parameter (list of ParamDef or name strings) so callers can pre-tick parameters. The picker now shows grouped views (All / Instance / Type). Same treatment for forms.select_family_parameters.
- ParamDef promoted to TemplateListItem: Fully backwards-compatible — supports the same attribute access, iteration, _fields, and _asdict() as the namedtuple it replaces, plus the new checked kwarg for preselection.
- Bug: duplicate results from grouped SelectFromList: _get_all_ctx() yields every item from every group, so a ParamDef present in both "All Parameters" and "Instance Parameters" was returned twice. Fixed with an identity-deduplication pass, guarded by isinstance(…, list) so single-select callers are unaffected.
- Bug: value column showed parameter name: PropKeyValue.__init__ fell back to name when display_value=None. Changed fallback to value, and get_source_properties now always captures AsValueString() as the display value.
- Icons extracted: All icon path data and the ClearButton style live in clipboard.Icons.xaml, loaded programmatically before wpf.LoadComponent (WPF loose-XAML has no base-URI context, so <ResourceDictionary Source="…"/> in the file itself would fail).

https://github.com/user-attachments/assets/3aedf4c8-ff82-404f-982e-78fdf55e9925

---

## Checklist

  ```bash
  ```

---

## Related Issues

- Resolves #2977
- Resolves #2840


jmcouffin:

PR #3327, opened 2026-04-23
fix: resolve NameError 'forms' in events.py (fixes #3285)

MayconAlvesss:
### Description
This PR resolves a `NameError: global name 'forms' is not defined` that occurs in `pyrevit/revit/events.py` when certain tools (like Measure 3D) are executed for the first time in a clean session. 

### Details
Added missing `from pyrevit import forms` import to ensure the forms module is available within the events handler context.

Fixes #3285

PR #3360, opened 2026-05-04
fix: restore module globals in execute_in_revit_context before ExternalEvent fires

### LINKED ISSUE #3285: [Bug]: On the first run of "Measure 3D" an error is thrown
### ✈ Pre-Flight checks

### 🐞 Describe the bug

after you run with a clean engine ctrl+alt+shift it works fine. loader issue?

### ⌨ Error/Debug Message

```shell
ERROR [pyrevit.revit.events] ExternalEvent error: name 'forms' is not defined
```

### ♻️ To Reproduce

_No response_

### ⏲️ Expected behavior

_No response_

```shell
6.4.0
```

### Additional context

_No response_

Wurschdhaud:
When the IronPython engine clears a non-persistent script scope, or when the cached engine for a command is the startup-script engine (same TypeId via SessionUUID:EngineType:Extension), module references like `forms`, `revit`, and any other split-backend module can disappear from `func.__globals__` before the ExternalEvent callback executes, causing NameError.

Snapshot all module-type globals at scheduling time and restore any that are missing just before the function is called. Only module objects (type(v) is type(compat)) are restored, so mutable Revit state (doc, uidoc, etc.) is never overwritten with stale values.

**Unsure, but seems to be a less of a bandaid fix than #3327. Maybe another AI has an opinion on that?**

---

## Checklist

  ```bash
  ```

---

## Related Issues

- Resolves #3285

### DIFF
--- a/pyrevitlib/pyrevit/revit/events.py
+++ b/pyrevitlib/pyrevit/revit/events.py
@@ -1,5 +1,6 @@
 """Revit events handler management."""
 #pylint: disable=unused-argument
+from six.moves import queue
 from pyrevit import HOST_APP
 from pyrevit import EXEC_PARAMS, DB, UI
 from pyrevit import framework
@@ -143,19 +144,25 @@ def stop_events():
 
 class _GenericExternalEventHandler(UI.IExternalEventHandler):
     def __init__(self):
-        self.func = None
-        self.args = ()
-        self.kwargs = {}
+        self._queue = queue.Queue()
 
     def Execute(self, uiapp):
         try:
-            if self.func:
-                self.func(*self.args, **self.kwargs)
-        except Exception as ex:
-            mlogger.error("ExternalEvent error: {}".format(ex))
+            while True:
+                fn = self._queue.get_nowait()
+                try:
+                    fn()
+                except Exception as ex:
+                    mlogger.error("ExternalEvent error: {}".format(ex))
+        except queue.Empty:
+            pass
 
     def GetName(self):
         return "GenericExternalEventHandler"
+
+    def schedule(self, func):
+        self._queue.put(func)
+
 
 if compat.IRONPY:
     _HANDLER = _GenericExternalEventHandler()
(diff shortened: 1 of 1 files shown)


pyrevitlib/pyrevit/revit/events.py
```diff
@@ -143,20 +144,26 @@ def stop_events():
 
 class _GenericExternalEventHandler(UI.IExternalEventHandler):
     def __init__(self):
-        self.func = None
-        self.args = ()
-        self.kwargs = {}
+        self._queue = queue.Queue()
```
sanzoghenzo: does the deque suggested by the AI not work? 
It doesn't need six, and it's easy as `append` and `popleft` (for FIFO) or `pop` (for LIFO, but for that a list is enough).
`pop` will raise an `IndexError` of empty

pyrevitlib/pyrevit/revit/events.py
```diff
@@ -143,20 +144,26 @@ def stop_events():
...
-        except Exception as ex:
-            mlogger.error("ExternalEvent error: {}".format(ex))
+            while True:
+                fn = self._queue.get_nowait()
+                try:
+                    fn()
+                except Exception as ex:
+                    mlogger.error("ExternalEvent error: {}".format(ex))
+        except queue.Empty:
+            pass
```
sanzoghenzo: with a deque you can just `while self._queue:` to loop until it is empty, no need to use the outer try/except

sanzoghenzo (commented):
I haven't used IronPython for far too much to know if my comments are applicable or not 😅 
looks good overall

pyrevitlib/pyrevit/revit/events.py
Wurschdhaud: I tried to outsmart the AI by reading up the differences between queue/deque on stackoverflow. Gonna try your suggestion.

sanzoghenzo (approved):

jmcouffin (approved):

jmcouffin:

PR #3396, opened 2026-05-26
chore(workflows): drop setup-dotnet, refine WIP artifact, document vendored deps refresh

jmcouffin:
Drops `actions/setup-dotnet` from `ci.yml`/`wip.yml`/`release.yml` (the `windows-2025` image preinstalls .NET 4.8 + 8 + 10, which covers every TFM the labs/engines/runtime/telem/autocmp builds need); takes `labs.build_deps` off the CI hot path by treating `dev/libs/{netfx,netcore}` as vendored DLLs, and exposes a new `pipenv run pyrevit build deps` CLI command (plus a `release/RELEASING.md` section) for maintainers to refresh those locally when bumping a `dev/modules/` submodule — the only build that still needed the EOL .NET Core 3.1 SDK; renames the WIP artifact to `pyrevit-wip-installers-<version>` (via a new `Get Install Version` step) and excludes `dist/**/*.wixpdb` from the upload; also SHA-pins `microsoft/setup-msbuild` in `ci.yml` to match the other workflows.


jmcouffin:

PR #3397, opened 2026-05-26
feat(loader): opt-out of inline script metadata parsing for faster startup

romangolev:
**Summary**

  Speeds up pyRevit bundle scanning / cold start by reducing filesystem syscalls and
  allocations in pyRevitExtensionParser, and adds an opt-in toggle that skips reading
  inline dunder metadata from script.py files entirely. A new Migrate Bundle dev tool helps
  extension authors lift inline __title__ / __author__ / etc. into bundle.yaml so they can
  safely flip the toggle off.

  **Changes**

  New setting: read_script_metadata (default: true)
  - core.read_script_metadata config flag
  (pyRevitLabs.PyRevit.PyRevitConsts.ConfigsReadScriptMetadataKey) — when false,
  pyRevitExtensionParser skips reading __title__, __author__, etc. from .py files and uses
  only bundle.yaml values.
  - Exposed in the Settings smartbutton UI with localized strings for
  en/de/es/fr/pt-br/ru/zh-CN.
  - Surfaced in Python via userconfig.PyRevitConfig.read_script_metadata.

  **Parser perf (ExtensionParser.cs)**
  - ParseExtension bails before the recursive component walk when the extension is disabled
  in pyRevit_config.ini; extConfig is hoisted once and reused.
  - ParseComponents pulls a single * listing per bundle dir and filters in memory for
  script/config/content patterns, collapsing multiple Directory.GetFiles syscalls into one
  per bundle.
  - ReadPythonScriptConstants uses File.ReadLines + enumerator-based
  ExtractPythonMultilineString, eliminating the allLines.Skip(lineIndex+1).ToList()
  allocation per multi-line dunder.
  - Localized title/tooltip/helpurl dicts are only allocated when there is actual localized
  data to merge (with a capacity hint).
  - ReadScriptMetadata flag is cached per parse pass to avoid per-script INI lookups.

  **New tool: pyRevitDevTools → Migrate Bundle**
  - IronPython pushbutton that walks an .extension, reads inline dunder metadata via the new
  parser, and merges it into each bundle's bundle.yaml. Has settings for merge policy and
  post-migration script cleanup, so authors can adopt the new toggle without losing
  metadata.
  
  **Test plan**

  - [ ] Toggle Read script metadata off in Settings → restart Revit → confirm bundles whose
  metadata lives in bundle.yaml still render correctly; bundles relying solely on inline
  dunders fall back to bundle defaults.
  - [ ] Toggle back on → confirm inline __title__ / __author__ / __doc__ / localized
  variants still appear.
  - [ ] Disable an extension via pyRevit_config.ini → confirm it's skipped (no component
  walk) and cold start is faster.
  - [ ] Run Migrate Bundle against a sample .extension, verify bundle.yaml files gain the
  expected fields per the chosen merge policy and that script-cleanup behaves as
  configured.
  - [ ] Sanity-check cold start time on a large extension set with the flag both on and
  off.


extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer Tools.panel/Migrate Bundle.pushbutton/script.py
```diff
@@ -0,0 +1,289 @@
+"""Migrate inline script dunders into bundle.yaml across an .extension folder.
+
+Uses pyRevitExtensionParser.ExtensionParser.ReadScriptMetadata to read raw
+dunder values from each bundle script, then writes/merges them into the
+bundle.yaml beside the script. Picks merge behaviour and optional script
+cleanup at runtime. Backs up any existing bundle.yaml as bundle.yaml.bak
+before overwriting.
+
+Runs on IronPython (pyrevit.forms.pick_folder is not available under CPython).
```
sanzoghenzo: Document _what_ it does, not _how_.
Implementation details can change, so documentation may drift from the code if not updated contextually. Remove all these lines or rephrase.
Also, what does "picks merge behaviour and optional script before overwriting" means?

@jmcouffin can we add an agent rule to copilot or devloai to catch this?

extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer Tools.panel/Migrate Bundle.pushbutton/script.py
```diff
@@ -0,0 +1,289 @@
...
+DUNDER_NAMES = (
+    '__title__', '__authors__', '__author__', '__doc__', '__helpurl__',
+    '__context__', '__highlight__', '__min_revit_ver__', '__max_revit_ver__',
+    '__beta__', '__cleanengine__', '__fullframeengine__', '__persistentengine__',
+)
+
+output = script.get_output()
+
+
+def clr_to_python(value):
```
sanzoghenzo, line 38: Note for the future: this could be a nice addition to the pyrevitlib

extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer Tools.panel/Migrate Bundle.pushbutton/script.py
```diff
@@ -0,0 +1,289 @@
...
+        if not choice:
+            return existing, False  # cancel => skip this bundle entirely
+        if choice.startswith('Use SCRIPT'):
+            merged[key] = new_value
+            changed = True
+
+    return merged, changed
+
+
+def comment_out_dunders(script_path):
```
sanzoghenzo: Just thinking out loud: isn't it better to parse the code with ast? That way we could also check if the dunder variables are used throughout the script.
But I admit I never used that, so it might be a ton of work for little gain...

Also, can we add the option to remove the line instead of commenting it?

extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer Tools.panel/Migrate Bundle.pushbutton/script.py
```diff
@@ -0,0 +1,289 @@
...
+    merged, changed = merge_yaml(existing, incoming, policy, rel)
+    if not changed:
+        skipped_no_change += 1
+        continue
+
+    if op.isfile(bundle_yaml):
+        try:
+            shutil.copy2(bundle_yaml, bundle_yaml + '.bak')
+        except Exception as ex:
+            output.print_md('- WARN backup failed for `{}`: `{}`'.format(rel, ex))
```
sanzoghenzo: Wouldn't it be simpler to display a warning at the beginning to remind the user to take a backup or commit the changes before running this? That way we avoid the problem with failed backups.

dev/pyRevitLoader/pyRevitExtensionParser/ExtensionParser.cs
```diff
@@ -396,6 +408,13 @@ private static ParsedExtension ParseExtension(string extDir, int revitYear = 0)
         {
             var extName = Path.GetFileNameWithoutExtension(extDir);
 
+            // Bail out before we walk the component tree if the user has disabled
+            // this extension in pyRevit_config.ini. extConfig is reused below so
+            // ParseExtensionByName runs once, not twice.
```
sanzoghenzo: Is this AI trying to show off how smart it is? :rofl:

Reword to something like "skip the component tree recursion if the extension is disabled" or remove it altogether

dev/pyRevitLoader/pyRevitExtensionParser/ExtensionParser.cs
```diff
@@ -513,10 +532,8 @@ private static ParsedExtension ParseExtension(string extDir, int revitYear = 0)
                 extensionTemplates.Count > 0 ? extensionTemplates : null,
                 revitYear);
 
-            // Read extension config from pyRevit config file (cached).
-            // Config is keyed by folder name (e.g. [extension_test.extension]) so it matches install and Python.
-            var config = GetConfig();
-            var extConfig = config.ParseExtensionByName(extName);
+            // extConfig was already loaded at the top of this method for the early
+            // disabled-extension bail-out; reuse it here instead of re-fetching.
```
sanzoghenzo: Remove AI smartass-ness

sanzoghenzo (changes requested):

extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer Tools.panel/Migrate Bundle.pushbutton/script.py
jmcouffin: @sanzoghenzo done for copilot https://github.com/pyrevitlabs/pyRevit/blob/develop/.github/copilot-instructions.md

dev/pyRevitLoader/pyRevitExtensionParser/ExtensionParser.cs
jmcouffin: @copilot  fix this

dev/pyRevitLoader/pyRevitExtensionParser/ExtensionParser.cs
jmcouffin: @copilot fix this

extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer Tools.panel/Migrate Bundle.pushbutton/script.py
jmcouffin: @copilot fix this

sanzoghenzo:
Note that I only got to line 500 or so of the extension parser, I'll get back to it whenever I have some time

jmcouffin (dismissed):

jmcouffin:
@romangolev I tweak quite a few things here and there. wired the settings switch, reviewed the migration tool flow, ...
tested the bundle migration and ended up applying for all builtin extensions.
I did not have enough time to really benchmark the switch
It looks fine
Once I get a green light from @sanzoghenzo  we are a go 
Good job!

sanzoghenzo:
Too many changes in the binaries, even GitHub mobile app can't take all those DLLs 🤣

jmcouffin:
> Too many changes in the binaries, even GitHub mobile app can't take all those DLLs 🤣

I know, bad practice to host the dlls in the repo...
We need to find a way to maintain the clone feature while removing the need to host the dlls...

dev/pyRevitLabs/pyRevitCLIAutoComplete/pyrevit-autocomplete.go
sanzoghenzo, line 1: @jmcouffin 
It's not the first time that I see changes to this file that just change the order of the flags... why?
This shouldn't come up in this PR (and any other PRs for that matter 😅 )

extensions/pyRevitCore.extension/pyRevit.tab/pyRevit.panel/Settings.smartbutton/script.py
```diff
@@ -831,6 +835,8 @@ def update_tab_previews(self):
 
     # save configs
     def _save_core_options(self):
+        request_reload = False
```
sanzoghenzo: move this closer to where it is used

extensions/pyRevitCore.extension/pyRevit.tab/pyRevit.panel/Settings.smartbutton/script.py
```diff
@@ -862,14 +868,42 @@ def _save_core_options(self):
             user_config.min_host_drivefreespace = 0
 
         user_config.load_beta = self.loadbetatools_cb.IsChecked
+
+        if (
+            self.new_loader.IsChecked != self._initial_new_loader
+            and not self.reload_requested
+            and not request_reload
```
sanzoghenzo: since this is initialized at line 838 and never modified, this check is pretty useless. remove

extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer Tools.panel/Migrate Bundle.pushbutton/script.py
sanzoghenzo, line 1: uhm, I thought ast would ease the line detection, not make it worse to know at which line a statement starts and ends 😅 

Also: if a script has a syntax error, is it worth it to fallback to regex? I would throw an error to please fix the script instead of trying to handle it, potentially causing more damage.
Thinking out loud here...

dev/pyRevitLoader/pyRevitExtensionParser/ExtensionParser.cs
```diff
@@ -1216,34 +1225,36 @@ private static List<ParsedComponent> ParseComponents(
...
                 }
 
-                // Merge localized values: bundle takes precedence over script
-                var finalLocalizedTitles = scriptLocalizedTitles ?? new Dictionary<string, string>();
-                var finalLocalizedTooltips = scriptLocalizedTooltips ?? new Dictionary<string, string>();
-                var finalLocalizedHelpUrls = scriptLocalizedHelpUrls ?? new Dictionary<string, string>();
+                // Merge localized values: bundle takes precedence over script. Keep
+                // the locals null until there's actually something to store so the
+                // common case (no localized dunders, no localized yaml keys) does
+                // not allocate three empty dictionaries per component.
```
sanzoghenzo: Isn't this little memory gain in front of a more complicated/error prone code?

sanzoghenzo (commented):
My cognitive load budget has expired 🤣 

The following are my rants, along with this: if a code needs that much comments about how it works, it is usually a bad sign and it needs to be refactored in smaller, more manageable and understandable chunks.

It it is just the AI commenting how good it is at faking reasoning, then I'm out.

extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer Tools.panel/Migrate Bundle.pushbutton/script.py
romangolev: it makes sence, thanks

extensions/pyRevitDevTools.extension/pyRevitDev.tab/Developer Tools.panel/Migrate Bundle.pushbutton/script.py
romangolev: that's completely true, you have a point here!

romangolev:
@sanzoghenzo got some fixes sent down the branch, should be good to go
@jmcouffin also merged develop so it does not have conflict 

should be ready to go, if not, ping me back 🙌

sanzoghenzo (approved):
Looks good!

jmcouffin:

PR #3400, opened 2026-05-27
Enhance release workflow with output handling and permissions

jmcouffin:
Added output for the release URL in the release job and updated permissions for the notify job to allow writing to issues. Adjusted the notify step to use the new release URL output for better integration.


jmcouffin:

PR #3442, opened 2026-06-20
Fix new logger default styling and emoji

ChrisCrosley:
# Make the Logger Pretty Again 

## Description

A little weekend cleanup. The recent updates to the logger broke the pyRevit default styling, so it was falling back to plain html. Added a catch to fall back to default styling when the user has not configured custom styling.

The Emojize regex allowed short codes with whitespace. For logs using colons in the text like:
`INFO [pyrevit.loader.sessionmgr] Load time: 23.8264694214 seconds :thumbs_up:`
Emojize would see `: 23.8264694214 seconds :` instead of skipping the first colon to find `:thumbs_up:`

---

## Checklist

  ```bash
  ```

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -580,11 +580,12 @@
         },
         "pymdown-extensions": {
             "hashes": [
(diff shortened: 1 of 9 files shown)


jmcouffin:

PR #3445, opened 2026-06-22
Update pyrevitlib\pyrevit\revit\db\query.py

Denver-22:
In get_families() families were duplicated by the number of FamilySymbols in them.

### DIFF
--- a/pyrevitlib/pyrevit/revit/db/query.py
+++ b/pyrevitlib/pyrevit/revit/db/query.py
@@ -687,7 +687,7 @@ def get_families(doc=None, only_editable=True):
     }.values()
     if only_editable:
         return [x for x in families if x.IsEditable]
-    return families
+    return list(families)
 
 
 def get_noteblock_families(doc=None):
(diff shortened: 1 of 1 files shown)


Denver-22:
@jmcouffin 
(I hope I can describe my proposal here.)
I fixed the error in this PR. But the function itself isn't very logically sound.
It's better to search for families using a direct collector.
I tested the differences between the functions. The proposed version differs in the following ways:
- finds more system families (because it retrieves them directly);
- runs 2-4 times faster.
```python
def get_families(doc=None, only_editable=True):
    """
    Retrieves a list of families from the given Revit document.

    Args:
        doc (Document, optional): The Revit document to retrieve families from.
                                  If not provided, defaults to DOCS.doc.
        only_editable (bool, optional): If True, only returns families that are editable.
                                        Defaults to True.

    Returns:
        list: A list of Family objects from the document. If only_editable is True,
              only includes families that are editable.
    """
    families = DB.FilteredElementCollector(doc) \
        .OfClass(DB.Family) \
        .ToElements()

    if only_editable:
        families = [
            fam
            for fam in families
            if fam.IsEditable
        ]

    return list(families)
```

jmcouffin:
> @jmcouffin (I hope I can describe my proposal here.) I fixed the error in this PR. But the function itself isn't very logically sound. It's better to search for families using a direct collector. I tested the differences between the functions. The proposed version differs in the following ways:

as long as the input and output are not breaking anything, I am ok with the improvements @Denver-22

PR #3474, opened 2026-07-06
Fix/config read telemetry path

thumDer:
# Fix/config read telemetry path

## Description

Telemetry path / url getters were wrong like always, returning telemetry flags. Now returning correct values.

---

## Checklist

  ```bash
  ```

---

## Related Issues

-


jmcouffin (approved):
Thanks @thumDer

PR #3487, opened 2026-07-14
Add `pyrevit.revit.avf` — helpers for displaying values via the Analysis Visualization Framework

Wurschdhaud:
### Summary

Adds a small helper module for Revit's AVF / `SpatialFieldManager` API, so
extension authors can paint arbitrary numeric values (KPIs, QC results,
heatmaps, counts, whatever) onto elements in a view without hand-rolling
SFM/schema/primitive bookkeeping every time.

This started as an internal tool for diagnostic/KPI overlays and I'm
proposing a generalized version for the shared lib.

### What's included

- `get_face_facing_direction(element, direction)` — returns the largest
  face of an element whose normal faces a given direction (typically
  `view.ViewDirection`). Used to pick a sensible point to anchor a marker.
- `display_avf_values(element_value_dict, view, ...)` — paints a dict of
  `{element: value}` onto a view in one call; creates/reuses the
  `SpatialFieldManager` and schema, returns primitive handles for later
  updates.
- `update_avf_values(primitive_dict, element_value_dict, view, ...)` —
  cheap re-update of existing primitives (no geometry re-query).
- `display_avf_values_at_points(point_value_pairs, view, ...)` — same idea
  but for arbitrary XYZ points with no host element (e.g. a sampled grid);
  automatically chunks into multiple primitives per Autodesk's documented
  500-points-per-primitive guidance.
- `remove_avf_primitive(view, sfp_id)` / `clear_avf_results(view)` —
  cleanup helpers.
- `get_or_create_marker_display_style(doc, style_name)` — get-or-create an
  `AnalysisDisplayStyle` configured to show text markers.
- `get_or_create_colored_surface_display_style(doc, style_name, ...)` —
  get-or-create the gradient/heatmap style instead, with min/max color,
  legend steps, and rounding exposed.

### Known limitations

- Non-planar faces (cylinders, etc.) are sampled at the UV-bounding-box
  center for marker placement, which is an approximation, not a true
  centroid.
- `use_bbox_center=True` is the fallback for elements without usable
  solid faces, but it still requires the element to have geometry
  visible in the target view.
- `display_avf_values_at_points` chunks at 500 points/primitive per
  Autodesk's documented recommendation for the no-geometry overload —
  not independently benchmarked here, just carried over from the docs.

---

## Checklist

  ```bash
  ```


jmcouffin (approved):
Very good contribution. You need to demo the usage publicly, it has a lot of potential use cases. @Wurschdhaud

jmcouffin:
@vinnividivicci pretty sure this will interest you!

PR #3521, opened 2026-07-30
Update get_name() in query.py for universal search

Denver-22:
Previously, element.Name was added only for DB.Workset. Now, a need has been identified for DB.Category as well.
It's clear that a universal solution is needed, based on searching for the "Name" attribute.
At the same time, priority is given to checking PY3.


jmcouffin (dismissed):

Denver-22:
I hope I didn't ruin anything with the last commit.

jmcouffin (approved):

PR #3542, opened 2026-08-09
Add horizontal auto-fit for legend columns

Wurschdhaud:
## Auto-fit column widths in Filter Legend rows

Mirrors the existing vertical auto-fit (row height from measured `TextNote` bounding boxes) but for the horizontal axis.

**Problem:** Column widths were fixed from config (`text_column_width`), so long filter names / parameter values / display values could overflow into the next column.

**Fix:**
- `text_note_horizontal_extent()` – measures a note's real left/right/width via its bounding box (mirrors `text_note_vertical_extent`).
- `autofit_column_widths()` – run once per legend view after all rows (header + data) are created, since column N's offset depends on column N-1's actual widest note. Computes per-column max width, expands any column that overflows its configured width (+`min_gap`), and translates the affected notes with `ElementTransformUtils.MoveElement` — no delete/recreate, so per-instance state is preserved.
- `script.py` now collects notes per column (`name_notes`, `param_notes`, `value_notes`) as rows are created, and calls `autofit_column_widths` right before appending the finished legend view.

Only widens columns, never shrinks below configured width. Swatch column excluded (fixed width, not text-driven).

---

## Checklist

  ```bash
  ```


romangolev:
works well 👌
tested with Revit 2024, 2027

romangolev (approved):

PR #3575, opened 2026-08-18
Fix/lib extension root on startup search paths

OnePowerUser88:
# Put library-extension roots on startup, hook, and command search paths

## Description

Python `extensionmgr` adds each library extension's directory (`Foo.lib/`) to `sys.path`. C# `new_loader` did not: startup and hooks only got nested `Foo.lib/lib/`, so packages at the `.lib` root (`Foo.lib/Pkg`) were invisible to `startup.py` and hooks. Command assemblies already had the root (Python-parity) but not nested `lib/`.

This PR uses one helper for startup, hooks, and command-type generation: add the `.lib` root, then nested `lib/` when it exists. Nested `lib/` is compatibility with the old C# startup search path, not Python-parity. A normal `Foo.lib/Pkg` layout is unchanged (`Directory.Exists` on nested `lib/`).

Rocket-mode cache was keyed on assembly version only (`3.4.2.0`). Command search paths are compiled into those assemblies, so enabling a new `.lib` or changing this generator did not rebuild them. The cache seed now includes the library-extension paths and a `CommandTypeGeneratorSchema` (`lib-root-1`). Bump the schema when codegen changes without a pyRevit version bump. DLL write time is still not hashed (that recompiled every extension on every rebuild).

---

## Checklist

  ```bash
  ```

---

## Related Issues

- N/A (no linked issue)

---

## Additional Notes

C# only — PEP 8 / Black do not apply. No Python `module_paths` change; the C# orchestrator owns startup and hooks after the entry-module refactor.

Tests: `LibraryExtensionSearchPathTests` + `HookManagerTests.BuildHookSearchPaths`.

Rocket-mode users pick this up on next Revit start (new command-assembly hash). No manual cache delete.

---

### DIFF
--- a/dev/pyRevitLoader/pyRevitAssemblyBuilder/AssemblyMaker/AssemblyBuilderService.cs
+++ b/dev/pyRevitLoader/pyRevitAssemblyBuilder/AssemblyMaker/AssemblyBuilderService.cs
@@ -76,13 +76,15 @@ namespace pyRevitAssemblyBuilder.AssemblyMaker
             LoadExtensionModules(extension);
 
             // Include generation-time inputs that affect emitted command types, so cache invalidates
-            // when runtime behavior changes (e.g. rocket mode toggles or loader binary updates).
+            // when runtime behavior changes (rocket mode, generator schema, or library-extension
+            // search paths — those paths are compiled into each command).
             string strategySeed = string.Join("|",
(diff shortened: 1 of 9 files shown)


romangolev (approved):
That's a good one, thanks! LGTM and tested with Revit 2026

PR #3577, opened 2026-08-19
docs: agentic coding rules

romangolev:
## Summary

Unifies the three scattered agent instruction files (`.github/copilot-instructions.md`, `CLAUDE.md`, `AGENTS.md`) into a single canonical `AGENTS.md` at the repo root. The vendor files become 3-line pointers to the canonical file.

## Changes

- **Inline comments** — strict allow-list: non-obvious _why_, warnings about invariants, tooling pragmas, and `TODO` / `FIXME` with an owner. Anything else (restating code, narrating the implementation, section banners) is removed by default.
- **Docstrings** — Google-style on public Python symbols and XML `///` on public C# members. Docstrings carry purpose, contract, side effects, and invariants — never signature restatements or implementation narration.
- **Centralization** — `.github/copilot-instructions.md` and `CLAUDE.md` are now 3-line wrappers pointing to `AGENTS.md`.
- **Version alignment** — "Supported Revit versions" floor corrected to Revit 2021+ (matches the C# loader requirement) and runtime targets aligned with the actual project files: 2021–2024 → `net48`, 2025–2026 → `net8.0-windows`, 2027+ → `net10.0-windows`.

## Migration

No code or runtime changes. Any agent that previously read `.github/copilot-instructions.md` or `CLAUDE.md` still discovers the rules via the new one-line pointers.


AGENTS.md
```diff
@@ -1,7 +1,289 @@
 # Agent Instructions
 
-This repository uses unified agent instructions located at:
+Unified agent rules for the pyRevit repository. Every AI assistant touching this repo (GitHub Copilot, Claude Code, opencode, Cursor, Aider, etc.) must follow this file.
```
sanzoghenzo: No need to state the obvious, agents already put this on the context. Let's keep the token count low

```suggestion
Agent rules for the pyRevit repository.
```

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
 
-> [.github/copilot-instructions.md](.github/copilot-instructions.md)
+The vendor-specific files (`.github/copilot-instructions.md`, `CLAUDE.md`) are thin pointers to this file; they exist only so each tool discovers the rules at its expected filename.
```
sanzoghenzo: Remove, nobody cares about this info

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
 
-All AI assistants (GitHub Copilot, Claude, opencode, etc.) should follow the guidelines in that file.
```
sanzoghenzo, line 7: Remove, this is the default behaviour

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+---
+
+## Project overview
+
+pyRevit is a Rapid Application Development (RAD) environment for Autodesk Revit. It lets users write automation tools and add-ins in Python (IronPython 2.7.12 default, CPython 3.12.3, or IronPython 3.4.0), C#, or VB.NET. The project also ships a CLI utility for deployment and a telemetry server for usage tracking.
```
sanzoghenzo: I may have missed something the past year, does pyrevit let you create a separate add-in?

```suggestion
pyRevit is a Rapid Application Development (RAD) environment for Autodesk Revit. It lets users write automation tools in Python (IronPython 2.7.12 default, CPython 3.12.3, or IronPython 3.4.0), C#, or VB.NET. The project also ships a CLI utility for deployment and a telemetry server for usage tracking.
```

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+- **docs** — Documentation source for the website (mkdocs).
+- **extensions** — pyRevit extensions (tools visible in the Revit ribbon). `pyRevitCore.extension` builds the pyRevit ribbon tab; the rest are enabled via the Extensions button. `pyRevitDevTools` is handy for running tests and checking that pyRevit (and your changes) is working.
+- **extras** — Extra files that come in handy (icons, dark-mode generator).
+- **licenses** — Licenses of the included third-party projects.
+- **pyrevitlib** — pyRevit and related Python libraries; imported by user scripts to ease Revit API development.
+- **release** — Static assets needed to build the final product (pyRevit and pyRevit CLI installers) plus build artifacts and installer configurations.
+- **site-packages** — Third-party Python packages made available by pyRevit to the user. Given that the main Python engine is IronPython 2.7.12, packages here must be compatible with it.
+- **static** — Assets for the website and YouTube channels. You can ignore it.
+
+## Architecture overview
```
sanzoghenzo, line 15: Since this is something that is (or should be) present in the developer documentation, couldn't we just add a link to the file in the docs folder (and adjust it if needed)? This would result in less maintenance work to avoid documentation drift

CLAUDE.md
```diff
@@ -1,182 +1,3 @@
...
-- Revit 2027+: .NET 10.0
-
-## Git Workflow
-
-- `develop` branch: Active development (always start here)
-- `master` branch: Release material only
-- `docs` branch: Documentation website
-- Feature branches from `develop`, PRs back to `develop`
-- Run `git submodule update` after switching branches
+All agent rules for this repository live in [AGENTS.md](AGENTS.md). This file exists because Claude Code discovers instructions at this path; it is not the source of truth.
```
sanzoghenzo: Can we make claude.md a link to agents.md? It works on Linux, not sure how well this is handled on Windows.

In any case, remove the useless explanation 


```suggestion
All agent rules for this repository live in [AGENTS.md](AGENTS.md).
```

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+- Detects any modifier keys held at click time and adjusts behavior accordingly.
+- Runs the appropriate script (Python, C#, Dynamo, etc.) based on the button's configuration.
+
+The appropriate script engine is selected automatically based on the script type.
+
+## Languages and technologies
+
+- **Python**: IronPython 2.7.12 (default), CPython 3.12.3, IronPython 3.4.2.
+- **C#**: .NET Framework 4.8 (Revit 2021–2024), .NET 8.0 (Revit 2025–2026), .NET 10.0 (Revit 2027+).
```
sanzoghenzo: .net 10 will be used also by recent versions of 2025 and 2026, if I understood correctly.
Also, see the related comment below

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+$env:DOTNET_ENVIRONMENT = 'Production'
+dotnet run -c Release -- ci
+```
+
+Other pipeline modes: `pack`, `sign`, `publish`, `winget`, `notify`. See `build/README.md` for the full list, `Build__Channel` semantics, and CI gating.
+
+## Documentation
+
+- Main website: https://pyrevitlabs.io/
+- Technical docs: https://docs.pyrevitlabs.io/ (mkdocs, built from `docs/` folder)
```
sanzoghenzo: Since we have the docs folder already available in markdown format and the reference API included in the docstrings, it doesn't make sense to point the ai to the online html version.

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+- `Pipfile` — Python dependencies (requires Python 3.14).
+- `pyRevitfile` — Engine definitions and deployment profiles.
+- `pyproject.toml` — Ruff linting config (Google docstring convention).
+- `mkdocs.yml` — Documentation generation.
+- `.gitmodules` — Git submodules for dependencies.
+
+## Code style
+
+- Python: Google docstring convention, formatted with black, linted with ruff.
```
sanzoghenzo: "formatted" and "linted" seem to suggest an automation, but this is not (yet) enforced by pre-commit hooks or anything else; it would be better to clearly instruct the ai to run them.
(Also, ruff can format the code like black, we could reduce the dev dependencies... But that's for another pr)

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+# Call get_ext_root_dirs to retrieve the list of extension paths from user config,
+# then pass each path to extensionmgr to scan for UI extension manifests.
+extensions = get_all_extensions()
+```
+
+**Example — what to write instead:**
+
+```python
+# Collect all installed extensions visible to the current user before building the UI.
+extensions = get_all_extensions()
```
sanzoghenzo: Not sure if this is the right example, this seems more appropriate for the get_all_extensions docstring (excluding the "before building the UI", which should be obvious in a subsequent line when "build_ui" is called).

Here the function name is self-explanatory, there's no need for a comment.

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+Code tells the agent **how**. Docstrings tell it **what, why, and what must remain true**.
+
+### Scope hierarchy
+
+Information lives in the narrowest scope that carries it:
+
+| Scope | Carries |
+|---|---|
+| `AGENTS.md` | Architecture, conventions, commands every agent needs |
```
sanzoghenzo:
```suggestion
| `AGENTS.md` | Conventions, commands every agent needs |
```

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+- Revit 2021–2024: .NET Framework 4.8.
+- Revit 2025–2026: .NET 8.0 (Windows).
+- Revit 2027+: .NET 10.0 (Windows).
+
+## Git workflow
+
+- `develop` branch: Active development (always start here).
+- `master` branch: Release material only.
+- `docs` branch: Documentation website.
```
sanzoghenzo: Add a note that this shouldn't be touched, since it is handled by ci

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+1. Scan for inline comments. For each one, name the special case it falls under. If you cannot, delete it.
+2. Scan for new or changed public symbols. Confirm each one has a docstring (Python) or XML `///` doc (C#) only if it carries information that cannot be read off the code — purpose, contract, side effects, invariants. Strip anything that just restates the signature or narrates the implementation.
+
+## Supported Revit versions
+
+2021–2027, with separate builds per version. The C# loader requires Revit 2021+; the legacy pure-Python loader (which supported older versions) has been removed.
+
+- Revit 2021–2024: .NET Framework 4.8.
+- Revit 2025–2026: .NET 8.0 (Windows).
+- Revit 2027+: .NET 10.0 (Windows).
```
sanzoghenzo: Same note as above. Since it is specifically ed here, we can drop the versions information at the beginning

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+```bash
+pyrevit clones add dev <path-to-repo>
+pyrevit attach dev default --installed
+```
+
+## Development workflow
+
+1. Fork and clone the repository.
+2. Checkout `develop` (active development).
```
sanzoghenzo: Repository is already cloned once the agent sees this file.
Develop should be the starting point to create a new branch, not the branch to work on

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+dotnet run -c Debug -- ci
+
+# WIP-style stamping + product build (mirrors develop push on the main repo)
+$env:Build__Channel = 'wip'
+dotnet run -c Release -- ci
+
+# Release-style stamping + product build (mirrors master / tag CI on the main repo)
+$env:Build__Channel = 'release'
+$env:DOTNET_ENVIRONMENT = 'Production'
+dotnet run -c Release -- ci
```
sanzoghenzo: Does the llm need to know all of this? Just list what it needs to run to check a successful build

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+```
+
+Other pipeline modes: `pack`, `sign`, `publish`, `winget`, `notify`. See `build/README.md` for the full list, `Build__Channel` semantics, and CI gating.
```
sanzoghenzo: Same as above, if the agent doesn't need to run it, remove this line

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+## Documentation
+
+- Main website: https://pyrevitlabs.io/
```
sanzoghenzo: I don't think this is useful. A direct link to the notion website would be better, if the ai can read it

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+- Technical docs: https://docs.pyrevitlabs.io/ (mkdocs, built from `docs/` folder)
+
+```bash
+pipenv run docs                       # Build documentation (mkdocs)
+pipenv run check-docstrings           # Lint docstrings with ruff
+```
```
sanzoghenzo, line 56: Is this section about where to find documentation or how to build it? 
if the former, remove this block. If the latter, rephrase to make it clearer, and move it next to the other instructions about documentation.
For agent usage, it may be better to run `mkdocs build` with the `--strict` flag to fail the build on warnings, so that the llm will try to fix them.

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+## Testing
+
+To test in Revit:
+
+```bash
+pyrevit clones add dev <path-to-repo>
+pyrevit attach dev default --installed
+```
```
sanzoghenzo, line 65: Move this to the related point in the next section

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+3. Initialize submodules: `git submodule update --init --recursive`.
+4. Install dependencies: `pipenv install`.
+5. Build: `cd build && dotnet run -c Debug -- ci && cd ..`.
+6. Test in Revit by attaching the clone.
+
+For debugging C# code:
+
+1. Build in Debug mode.
+2. Open the appropriate `.sln` file in Visual Studio.
+3. Attach the debugger to `revit.exe`.
```
sanzoghenzo: Is this something the agent can do? If not, remove it.

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+MyExtension.extension/
+  MyTab.tab/
+    MyPanel.panel/
+      MyButton.pushbutton/
+        bundle.yaml      # Button configuration
+        script.py        # Python script
+        icon.png         # Button icon
+```
+
+Supported bundle types: `pushbutton`, `smartbutton`, `pulldown`, `splitbutton`, `panelbutton`.
```
sanzoghenzo: This seems to me a little out of.. context 😅 
It may be better to place it next to the project hierarchy explanation (if it is needed at all).

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+## Key configuration files
+
+- `Pipfile` — Python dependencies (requires Python 3.14).
+- `pyRevitfile` — Engine definitions and deployment profiles.
+- `pyproject.toml` — Ruff linting config (Google docstring convention).
+- `mkdocs.yml` — Documentation generation.
+- `.gitmodules` — Git submodules for dependencies.
```
sanzoghenzo: Does the agent need to know this?

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+## Code style
+
+- Python: Google docstring convention, formatted with black, linted with ruff.
+- C#: Standard .NET conventions.
+
+## Commenting guidelines
```
sanzoghenzo, line 89: Can we just say "don't write inline comments. Ever."? 😅 
Also "Code must be self-explanatory. Keep existing comments." Should resolve the matter with few words.
Maybe keep the pragma and utf-8 thing.

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+When applicable, capture:
+
+- **Purpose** — what abstraction or business operation this represents.
+- **Contract** — inputs, outputs, important guarantees.
+- **Side effects** — DB writes, network calls, filesystem changes, emitted events.
+- **Exceptions** — especially domain-specific ones and what they mean here.
+- **Invariants** — conditions that must remain true after refactoring.
+- **Non-obvious constraints** — ordering, idempotency, thread safety, transaction boundaries, compatibility requirements.
+- **Why something unusual exists** — historical or architectural reason.
```
sanzoghenzo: I am worried about this, I believe it can be read as an excuse by the LLM to keep citing previous behaviour

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+- **Architectural role** — how this symbol relates to the rest of the system (e.g. "Payment providers must not modify subscriptions directly").
+
+Mark dangerous-to-break constraints explicitly with Google-style sections: `Important:`, `Note:`, `Warning:`, `Invariant:`. Treat them as red lines for the next editor.
+
+### What does not belong
+
+- Restating the signature, name, or types.
+- Narrating the implementation (`"""Loop through items and add item.price to total."""`).
+- Padding trivial functions — if behaviour is obvious from the code, omit the docstring entirely.
```
sanzoghenzo, line 134: If the function is public (I'm thinking about pyrevitlib here), the docstring will make into the documentation site, while the code block will be collapsed, so it might be non obvious to understand it. I would drop this line

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+- Implementation details an agent can inspect (`"""Uses Redis sorted set keyed by timestamp."""` when the call is right there).
+
+### Language rules
+
+- **Python**: Google-style docstrings on **public** symbols (non-underscored). Match `pyproject.toml` (`select = ["D"]`, `convention = "google"`), enforced by `pipenv run check-docstrings`.
```
sanzoghenzo:
```suggestion
- **Python**: Google-style docstrings on **public** symbols (non-underscored).
```
The info on how to lint docstrings that were placed in the "Documentation" section should go here (maybe, I'm always confused reading LLM crafted skills and agents.md, they doesn't seem to have a clear structure and repeat things needlessly)

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+- **C#**: XML `///` doc comments on **public** classes and methods. `<summary>` carries purpose; `<param>` / `<returns>` / `<exception>` cover non-obvious details only.
+- Private (underscored Python, `private` / `internal` C#) members are exempt — same carve-out as `D107` and `D105`.
```
sanzoghenzo:
```suggestion
- Private (underscored Python, `private` / `internal` C#) members are exempt.
```

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+When behaviour changes, update the docstring / XML doc in the same change.
+
+## Self-check before finalising
+
+Before finishing a task, audit the diff:
+
+1. Scan for inline comments. For each one, name the special case it falls under. If you cannot, delete it.
```
sanzoghenzo: See note above, in my head there shouldn't be any special case... Just keeping existing ones (if made by a human and still valid)

AGENTS.md
```diff
@@ -1,7 +1,289 @@
...
+
+### Language rules
+
+- **Python**: Google-style docstrings on **public** symbols (non-underscored). Match `pyproject.toml` (`select = ["D"]`, `convention = "google"`), enforced by `pipenv run check-docstrings`.
+- **C#**: XML `///` doc comments on **public** classes and methods. `<summary>` carries purpose; `<param>` / `<returns>` / `<exception>` cover non-obvious details only.
+- Private (underscored Python, `private` / `internal` C#) members are exempt — same carve-out as `D107` and `D105`.
+
+When behaviour changes, update the docstring / XML doc in the same change.
+
+## Self-check before finalising
```
sanzoghenzo: I'm still a newbie regarding prompt engineering, but since we already stated the constraints, what's the point of this self check? Beside polluting the context with repeated info, that is.

sanzoghenzo (commented):
Here's my highly opinionated but unqualified review.

I'll only cite [this research](https://arxiv.org/html/2602.11988v1) in which they found that LLM generated agents.md worsen the coding tasks.

I don't use ai that much, and I still have to do proper evals, but I found that for my python projects a 20-30 lines agents.md with commands to check, format and test the code and brief instructions to avoid repeated bad behaviour goes a long way.

Now that I think of it, wouldn't be better to split this into files to put in the appropriate folders (for example python and docstring things are only useful for pyrevitlib, and - maybe the extensions folder)?

AGENTS.md
romangolev: we have extensions though, not add-ins exactly

AGENTS.md
romangolev: If we are going to touch this ground, I would also like to suggest some C# formatting as well. Together or separately with ruff

AGENTS.md
romangolev: removed from agents, added a pointer

AGENTS.md
romangolev: moved it as a separate description as a md file. In case somebody would like to build an extension with agent, it would be helpful

docs/architecture.md
sanzoghenzo: I believe this is not true anymore, is it?

docs/architecture.md
```diff
@@ -126,4 +130,10 @@ In turn, the code in `ScriptExecutor.cs` calls the appropriate script engine bas
 
 ???+ info
 
-    You can find the code of the engines in the files that end `Engine.cs`.
+    You can find the code of the engines in the files that end `Engine.cs`, in `dev/pyRevitLabs.PyRevit.Runtime/`:
```
sanzoghenzo:
```suggestion
    You can find the code of the engines in the files that end with `Engine.cs`, in `dev/pyRevitLabs.PyRevit.Runtime/`:
```

sanzoghenzo (commented):
Great, just a couple more things.

How should one test the effectiveness?
Just run the same task with or without the agents.md and check the best results?

AGENTS.md
```diff
@@ -1,7 +1,142 @@
...
+pyRevit is a Rapid Application Development (RAD) environment for Autodesk Revit. It lets users write automation tools in Python (IronPython 2.7.12 default, CPython 3.12.3, or IronPython 3.4.0), C#, or VB.NET. The project also ships a CLI utility for deployment and a telemetry server for usage tracking.
+
+## Architecture overview
+
+See [`docs/architecture.md`](docs/architecture.md) (also links to [`docs/extensions.md`](docs/extensions.md)).
+
+## Supported Revit versions
+
+- Revit 2021–2024: .NET Framework 4.8.
+- Revit 2025–2026: .NET 8.0 /.NET 10
```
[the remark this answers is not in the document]
jmcouffin: support for .NET10 is more of a consequence than a cause.

jmcouffin (approved):

PR #3559, opened 2026-08-14
Remove dead pre-2021 HOST_APP version checks

jmcouffin:
Follow-up to #3438 (drop Revit 2013–2020): unwrap Python host-version gates that are now always true or always false.

`HOST_APP.is_newer_than(v)` is `host > v`. With a minimum host of 2021, checks against 2013–2020 (and `is_newer_than(2021, or_equal=True)`) no longer branch. This PR keeps the remaining supported-API path and deletes the unused MatchConfigWindowLegacy UI.

## Description

Removes dead `HOST_APP.is_newer_than` / `is_older_than` / equivalent `HOST_APP.version` comparisons in `pyrevitlib` and bundled extensions. Live gates for 2022+, 2024+, and 2026+ are unchanged.

`is_newer_than(2021)` **without** `or_equal` is still live (false on 2021, true on 2022+). Those call sites were left as-is.

Also removes redundant `min_revit_version` keys from `bundle.yaml` files when the floor is 2014/2016/2019 (always satisfied on Revit 2021+). 2022+ minima and the empty test-bundle schema keys are unchanged.

## Checklist

- [x] Code follows the [PEP 8](https://peps.python.org/pep-0008/) style guide.
- [ ] Code has been formatted with [Black](https://github.com/psf/black) (style left matching surrounding files; IronPython 2.7).
- [ ] Changes are tested and verified to work as expected.

## Related Issues

Follow-up to #3438, as noted in https://github.com/pyrevitlabs/pyRevit/pull/3438#issuecomment-5289779932

## Additional Notes

Quoted related leftovers **not** in this PR (C# preprocessor, still-live Python gates):

- Runtime `#if REVIT2013` … `REVIT2020` branches in `dev/pyRevitLabs.PyRevit.Runtime/` (`EventTelemetry.cs`, `EventHandling.cs`, `EventHooks.cs`, `ScriptCommands.cs`, `ScriptConsoleUtils.cs`, `CLREngine.cs`, `ContentEngine.cs`, `GrasshopperEngine.cs`, plus `RevitThemeDetector.cs`). Those can be a separate cleanup.
- Remaining Python version gates that are still meaningful: `is_newer_than(2021)` (2022+), `2022` / `2022, or_equal=True`, `2024`, `2026`, `is_older_than(2023)`, `is_older_than(2027)`, `is_exactly(2021)` / `is_exactly('2022')`.
- Remaining `min_revit_version` values that still hide tools on older hosts: 2022, 2023, 2025, 2027.
[...]

PR #3585, opened 2026-08-25
Drop legacy pre-2021 Revit fallbacks (runtime + loader)

romangolev:
## Summary
- Removes the pre-2017 `UIApplication` field fallback (`NEW_UIAPP_FIELD_VERSION`/`LEGACY_UIAPP_FIELD`) from the C# loader, now that the minimum supported host is Revit 2021.
- Strips `REVIT2013`..`REVIT2020` preprocessor branches and their fallbacks across `pyRevitLabs.PyRevit.Runtime`, and tightens the loader theme detector's `UIThemeManager` gate to REVIT2021+.

Split out of #3559 per review feedback — these are C# preprocessor branches rather than the Python `HOST_APP` gates that PR targets.

## Test plan
- [ ] Build and attach a dev clone against a Revit 2021+ host, confirm session loads and dark-mode theme detection still works.


jmcouffin (approved):
lgtm

